PR #1, opened 2026-06-21
[docs] Add extension library foundation

franksong2702:
## Summary
- add the initial README for the Hermes WebUI extension library purpose, layout, and trust model
- add contribution rules and PR checklist for extension entries
- add a repository-side extension entry contract and minimal manifest-bundle example
- document sidecar metadata expectations for loopback helpers such as Desktop Companion
- add compatibility/testing guidance so existing extensions can be checked as the main WebUI extension contract evolves

## Why
This repo is new and needs a small shared convention before real extensions are submitted. The PR is docs-only so it can fit around the ongoing WebUI extension foundation work without adding Desktop Companion assets yet.

The latest update also captures the Discord consensus that sidecar extensions should be declared in manifest metadata and that this repo should eventually support compatibility/testing checks for existing extensions.

## Verification
- git diff --check
- python3 -m json.tool examples/manifest-bundle/manifest.json
- python3 -m json.tool examples/loopback-sidecar/manifest.json

## Notes
- No extension implementation is included
- No Desktop Companion code or assets are included
- Sidecar metadata is documented as descriptive until the main WebUI repo defines and ships the corresponding contract
- AI-assisted with Codex

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,38 @@
+# Contributing
+
+Thanks for helping build the Hermes WebUI extension ecosystem.
+
+The extension system is still taking shape, so keep PRs small and explicit.
+If an extension needs new WebUI APIs, start by describing the requirement
+before adding a large implementation.
+
+## Contribution Rules
+
+- Keep each extension self-contained under `extensions/<extension-id>/`.
+- Include a `README.md` with install, disable, uninstall, and trust notes.
+- Include a `manifest.json` that uses same-origin asset paths.
+- Do not commit secrets, tokens, credentials, cookies, private endpoints, or
+  machine-specific state.
+- Do not load remote third-party scripts at runtime.
+- Disclose any network, filesystem, native host, sidecar, or OS integration.
+- Bind sidecar services to localhost by default and document any ports used.
+- Prefer docs and examples when the WebUI extension contract is still changing.
+
+## Pull Request Checklist
+
+Before opening a PR, confirm:
+
+- [ ] The change belongs in the extension library, not Hermes WebUI core.
+- [ ] The extension or example is isolated under one directory.
+- [ ] The README explains the user workflow and trust model.
+- [ ] The manifest uses local extension assets.
+- [ ] Install, disable, and uninstall behavior is documented.
+- [ ] Any sidecar or native integration is documented clearly.
+- [ ] No secrets, generated local state, or unreviewed binaries are included.
+
+## Review Expectations
+
+Maintainers may ask for an extension to start as documentation or an example
+until the WebUI-side API is ready. This is expected while the foundation work
+lands in stages.
+
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -16,6 +16,8 @@ before adding a large implementation.
 - Do not load remote third-party scripts at runtime.
 - Disclose any network, filesystem, native host, sidecar, or OS integration.
 - Bind sidecar services to localhost by default and document any ports used.
+- Document the WebUI version or extension API surface the entry was tested
+  against.
 - Prefer docs and examples when the WebUI extension contract is still changing.
 
 ## Pull Request Checklist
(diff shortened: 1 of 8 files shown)


nesquena-hermes:
Reviewed this against the actual WebUI extension loader (`api/extensions.py`) and `docs/EXTENSIONS.md` on `main` of the app repo. Short version: this is a clean, accurate, docs-only foundation that matches the shipped contract — I verified the claims empirically rather than just reading them. A few non-blocking notes and one direction question below.

## What I verified

- **Both example manifests load correctly through the real loader.** I ran `examples/manifest-bundle/manifest.json` and `examples/loopback-sidecar/manifest.json` through `api.extensions.get_extension_config()` with a temp extension dir. Both resolve as expected: bare relative `assets/example-extension.js` → `/extensions/assets/example-extension.js`, matching `_manifest_asset_url()` exactly. The extra `id` / `name` / `sidecar` fields are silently ignored by `_iter_manifest_entries()` (which only reads `scripts`, `stylesheets`, `enabled`), so they're harmless forward-looking metadata — correct.
- **The `docs/EXTENSIONS.md` cross-reference (README line 15-16) is valid** — that file exists on `main` of the app repo, and its manifest shape, same-origin rule (`/extensions/` or `/static/` only), and loopback-by-default guidance are all consistent with what's documented here. No contradiction between the two repos.
- **Trust model is stated correctly.** "Extension JS runs in the WebUI origin and can interact with the authenticated session… review like application code, not passive themes" is exactly right and is the single most important thing for a curated library to get right. 👍

## Non-blocking suggestions

1. **`sidecar.origin` is descriptive only — worth making that even more unmissable in the example.** The loader's `_is_safe_asset_url()` rejects any URL with a scheme/netloc, so `http://127.0.0.1:17787` is never something WebUI fetches or injects today; it's pure metadata. `docs/extension-entry.md` says this ("descriptive until the main WebUI repo defines stronger behavior") and the example README repeats it, which is good. Consider a one-line comment-equivalent in the loopback example README explicitly stating "WebUI does not connect to this origin yet" so a contributor can't mistake it for an active capability.

2. **The `id` field has no loader meaning yet — say so.** `docs/extension-entry.md` presents `id` in the manifest as if load-bearing, but the loader ignores it (the directory name under `extensions/<extension-id>/` is what's canonical). Worth a sentence clarifying that `id` is a repo-side/forward-looking convention, not something the current loader keys on, so the two notions of "id" (manifest field vs directory) don't drift.

3. **`assets/` paths vs. the example layout.** `docs/extension-entry.md` shows `extensions/<extension-id>/assets/` and the manifest references `assets/example-extension.js`. Just flag that the deployed extension dir (`HERMES_WEBUI_EXTENSION_DIR`) is the resolution root, so when a real extension lands, the manifest's relative paths must be relative to *that* root, not to `extensions/<id>/`. A one-liner in `extension-entry.md` mapping repo layout → deployed layout would pre-empt the most likely first-contributor mistake.

4. **Tiny:** the loader caps at 32 URLs / 64KB manifest (`_MAX_URL_LIST`, `_MAX_MANIFEST_BYTES`). Not worth documenting per-extension, but if you ever add a CI validator here (the README mentions that as a future goal), those are the limits to check against.

## One direction question

The sidecar metadata (`type: loopback`, `origin`, `health_path`) is the only part of this PR that's ahead of the shipped WebUL contract — the app side has no sidecar field yet. The docs are appropriately careful to call it descriptive, so it's safe to merge as docs. My question is about sequencing: do we want the **canonical sidecar schema to be defined in the app repo's `docs/EXTENSIONS.md` first** (so this library mirrors it), or is it intentional to let the desired shape be proposed here and then implemented app-side? Either is fine; I just want to avoid the two repos shipping slightly different sidecar field names later. If the answer is "app repo is the source of truth," a one-line pointer here saying the sidecar schema is provisional and owned by the app repo would lock that in.

## Recommendation

**Mergeable as-is.** It's docs-only, accurate to the current loader, internally consistent, consistent with the app repo, and correctly scoped (no registry UI, no install flow, no Desktop Companion assets — matching the stated first-phase scope). None of the notes above are blocking; they're polish that could also land in a follow-up once the first real extension or the sidecar app-side contract arrives.

Nice foundation, @franksong2702 — verified end-to-end and it holds up.

PR #2, opened 2026-06-21
docs: add VISION.md — extensions system roadmap & direction

nesquena-hermes:
## Summary

Adds `VISION.md` — a public-facing vision & roadmap for the extension library, so contributors can build toward a shared target.

It captures the direction agreed in the founding maintainer discussion:

- **The thesis** — lean curated core; everything else becomes a vetted extension. Out-of-scope-for-core ideas get redirected here instead of rejected.
- **Two-repo split** — infrastructure/contract in `nesquena/hermes-webui`, entries/conventions/registry here.
- **Trust model** — extensions are trusted local code running in the WebUI origin; *being merged is the vetting*.
- **Capability ladder** — asset bundling ✅ → Settings UI (in progress) → sidecar metadata (foundation merged) → backend routes/proxy (planned: declarative per-extension opt-in same-origin proxy to a declared loopback sidecar first; in-process route handlers later and separately gated).
- **Registry & install target** — a GitHub Action regenerates a full `registry.json` on every merge; Settings → Extensions reads it; browse → Install → confirm → install in place → restart if required; manage/disable/uninstall from the same surface.
- **Compatibility & testing** — entries declare required capabilities; CI validates manifest/metadata on every PR.

Docs-only; links only to files already on `main` (`CONTRIBUTING.md`, `docs/extension-entry.md`). Companion tracking issues break the roadmap into concrete pieces.

It's a living plan, not a locked contract — happy to adjust framing before merge.

### DIFF
--- a/VISION.md
+++ b/VISION.md
@@ -1,0 +1,101 @@
+# Vision & Roadmap — Hermes WebUI Extensions
+
+This repository is the **curated extension library** for Hermes WebUI. This
+document describes where it's going so contributors can build toward a shared
+target. It's a living plan, not a frozen spec — expect it to evolve as the
+foundation lands in stages.
+
+## The thesis
+
+Hermes WebUI core stays **lean and carefully curated** — deliberately *not*
+kitchen-sink. Just because something *can* be built into the core for everyone
+doesn't mean it *should* be. Everything genuinely useful that doesn't belong in
+the curated core gets a first-class home as an **extension**.
+
+The goal: a large library of **vetted** extensions, and an in-WebUI
+**Settings → Extensions** experience where you browse a gallery and install what
+you want in one click. The core stays clean for everyone; a rich ecosystem grows
+around it. Out-of-scope-for-core ideas get *redirected here* instead of rejected.
+
+## Two repos, clean split
+
+- **`nesquena/hermes-webui` (core)** — the extension *infrastructure*: the loader,
+  manifest contract, same-origin asset injection, sandboxed static serving, the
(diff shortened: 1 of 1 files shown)


santastabber:
Only wording concern: the roadmap should distinguish the current extension sidecar pattern from the future proxy work.

Today, extension JS can communicate directly with trusted loopback sidecars when CSP allows it. A same-origin per-extension proxy is still planned/future core work, not something the repo entries can rely on yet.

Suggested tweak: phrase it as “planned future same-origin proxy support” or “future Tier A proxy work” rather than implying proxy support is already part of the shipped extension contract.

nesquena-hermes:
Completely right, and thanks for the precise correction — I had the proxy framed as more present-tense than it is.

I confirmed it against core: the CSP `connect-src` base already allows `http://127.0.0.1:*` / `localhost:*` / `ws://` loopback (`api/helpers.py` `_CSP_CONNECT_BASE`), so **direct browser → trusted loopback sidecar is the current shipped pattern** — no proxy required for it. The same-origin `/extensions/<id>/sidecar/*` proxy is future Tier A work for the cases the direct path can't cover (the sidecar's own CORS, hiding the port, uniform same-origin auth), and repo entries shouldn't depend on it yet.

Updated VISION.md to split capability #3 (sidecar metadata + **direct loopback, shipped**) from #4 (**future** opt-in same-origin proxy), and to state explicitly that sidecar extensions work today via direct loopback and entries must not assume the proxy. Pushed to this branch.

Issue #8, opened 2026-06-21
Safety gates: automated malicious-code scan + integrity + sanity checks on every entry PR


nesquena-hermes:
## Goal

Build the automated safety gates that protect the curated extension set as contribution volume grows. Multiple human maintainers can review/merge new-entry PRs (Nathan + Santastabber + Frank), but **automated GitHub Actions** are what make that safe and consistent — they're the thing standing between "anyone can open a PR" and "users install with one click."

## Why this is load-bearing

Extensions run **in the WebUI origin against the authenticated session**, and the one-click install (now SHIPPED — core v0.51.644 / #4746) trusts the registry wholesale (being merged == vetted). So the gates aren't nice-to-have — they're the trust boundary, and that boundary is now load-bearing in production rather than hypothetical.

## Planned gates (build toward; not all at once)

- [ ] **Malicious-code / intent scan** — static analysis of an entry's added JS/assets for dangerous shapes: obfuscation / `eval` / `Function`, credential or token access, outbound-network primitives or third-party-script loaders, attempts to read/exfiltrate the authenticated session, `fetch`/DOM patterns reaching beyond the extension's own surface. Fail-closed on anything suspicious; a human maintainer clears or rejects. (The core repo already runs an analogous PR-diff threat scanner — reuse that shape.) _Partial — first fail-closed layer LANDED via **#15** (merged 2026-06-24): `scripts/scan-extension-safety.mjs`, wired into CI before registry generation. It catches committed secrets (6 token shapes), symlinks/unsafe paths, high-risk JS (`eval`/`Function`, string-eval timers, `document.cookie`, `process.env`, `child_process`/`fs`, `Deno`/`Bun` exec, external dynamic `import()`, remote-`<script>` loaders), undeclared external-URL literals (vs `permissions.network_external`), and undeclared `localStorage` writes. **Verified empirically** — each gate fires on injected fixtures, incl. cases the #13 validator passes (e.g. `document.cookie`). Remaining: a deeper **adversarial** scan — the regex layer has known, by-design bypasses (split-string URLs `'http'+'s://…'`, aliased `Function` via `globalThis['Fun'+'ction']`, computed `import()`, `XMLHttpRequest`); document them in-code so a green scan is never read as a security guarantee._
[...]

Issue #6, opened 2026-06-21
CI: validate extension manifest/metadata on every PR (schema + structure + capabilities)


nesquena-hermes:
## Goal

Add CI that validates every extension entry on each PR, so the library stays consistent as it grows and the registry generator can trust its inputs. This is the cheap, high-value first step of the compatibility/testing layer.

## Scope (start small)

- [x] **Manifest/metadata schema validation** — validate each `extensions/<id>/extension.json` (and the loader `manifest.json`) against the JSON schema on every PR. Invalid entry fails the check. ✅ live in `extensions.yml` via #13 (`validate-extensions.mjs`).
- [x] **Structural checks** — required files present (`README.md`, manifest), asset paths are same-origin / repo-local, IDs are lowercase-hyphen and unique, no secrets/binaries committed. ✅ #13.
- [x] **Capability declaration** — confirm each entry declares the extension API surface / capabilities it requires (capability names, not just a version pin), so core can roll forward and maintainers can see what breaks. ✅ #13 (validates against known shipped capabilities + permission-vs-code drift).

## Later

- Compatibility checks against core releases (does an entry's declared capabilities still exist in the current core contract?).
- Re-validate on merge as part of the registry generator (the generator and the PR check share the schema).

## Notes

- Shares the JSON schema with the registry generator issue — build the schema once, use it in both the PR check and the generator.
- Keep it lightweight at first (schema + structure); grow toward capability/compat checks as real entries accumulate. Both maintainers flagged this as needed *before* the library fills up.

Part of the extensions roadmap (see VISION.md §compatibility-and-testing). Pairs with the registry generator issue.

nesquena-hermes:
## Progress update — a per-entry validator now exists as a working precedent; the generic repo-wide CI is still to build

Desktop Companion (PR #11) shipped `scripts/validate-desktop-companion.mjs`, which is a concrete, working example of the checks this issue wants — it validates the entry's `extension.json`/`manifest.json`/pet manifests, asserts required files + capabilities, and **cross-checks the declared `webui_api` read/write surface against the actual adapter code** (so disclosure drift fails validation). It's currently entry-specific and run manually, not wired into CI.

**Still open — generalize it into repo-wide PR CI:**
- [ ] A **generic** validator that runs over `extensions/*/` (not one bespoke script per entry) — schema-validate every `extension.json`, validate the loader `manifest.json` shape.
[...]

Issue #5, opened 2026-06-21
Backend routes for extensions: declarative per-extension loopback sidecar proxy (Tier A)


nesquena-hermes:
> **Status (2026-06-23):** deferred / demand-driven — direct browser→loopback already works (proven by Desktop Companion), so this is only for cases direct can't cover. Core-side implementation tracked in nesquena/hermes-webui#4747. Roadmap umbrella: #12.

## Goal

Unblock **richer extensions that need backend behavior** without giving extension code the WebUI server's full authority. The current wall: extensions can ship JS/CSS but can't add same-origin backend routes, so anything server-side resorts to an external DB + webhook + fighting CORS (e.g. the Templates extension), and sidecar extensions (Desktop Companion) have no clean way for their frontend to reach their local backend.

### Tier A — declarative loopback proxy (build this first)

- An extension **declares** a local sidecar in its metadata (the `sidecar` block already drafted: `{ type: "loopback", origin: "http://127.0.0.1:<port>", health_path }`).
- WebUI exposes a **same-origin, per-extension, allowlisted** proxy — e.g. `/extensions/<id>/sidecar/*` → the declared localhost origin.
- **Opt-in and user-confirmed at install time.** The extension's backend runs as a **separate localhost process the user controls** (Tauri shell / helper binary / local server), never inside the WebUI process.
- Net: the extension's frontend calls its own backend same-origin (no CORS), while the trust surface stays "a localhost process you chose to run." Directly unblocks Templates and Desktop Companion.

### Tier B — in-process route handlers (defer; separately gated)

- Extensions shipping server code that registers real WebUI routes is far more powerful but runs arbitrary code in the server process.
- Needs an explicit trust gate and likely subprocess isolation (cf. the core's subprocess-plugin explorations).
- **Defer** until a concrete need arises that Tier A can't meet; treat as its own security-boundary decision.

## Tasks (Tier A)

- [ ] Finalize the `sidecar` metadata contract in the core loader (today it's descriptive only — the loader rejects any scheme/netloc URL)
- [ ] Design the per-extension proxy route + allowlist model
- [ ] Install-time opt-in + user confirmation for proxy enablement
- [ ] Optional: coarse sidecar health via `health_path`
- [ ] Security review of the proxy boundary (localhost-only target, no SSRF, per-extension scoping)

## Notes

- Core-side work (`nesquena/hermes-webui`); this issue tracks direction.
[...]

Issue #4, opened 2026-06-21
Settings → Extensions UI: browse, install (confirm + restart-if-needed), disable/uninstall


nesquena-hermes:
## Goal

Add a **Settings → Extensions** surface in WebUI core (`nesquena/hermes-webui`) where users browse the curated library, install extensions, and manage installed ones. This is the user-facing payoff of the whole extension system.

## Target flow (from VISION.md)

1. **Browse** — a Settings → Extensions tab reads the generated `registry.json` and shows the curated extensions: name, description, screenshots, trust disclosures.
2. **Install** — open an entry → read about it → hit **Install** → confirmation dialog ("Are you sure you want to install this extension?") → install in place → **restart WebUI if the extension requires it** (entry declares `restart_required`).
3. **Manage** — the same surface lists installed extensions and lets the user **disable** or **uninstall** them.

## Phasing (each independently shippable)

- [x] **Phase 1 — read-only listing.** Render the registry; no install yet. ✅ shipped in core (v0.51.644 / nesquena/hermes-webui#4746) — Settings → Extensions → **Gallery** tab fetches the registry (proxied through `/api/extensions/registry`, 5-min TTL) and renders each entry with metadata, tags, capabilities, and permission disclosures.
- [x] **Phase 2 — enable/disable** of already-present extensions. ✅ shipped in core (v0.51.553 / #4637, `extension-overrides.json`, auth+CSRF-gated).
- [x] **Phase 3 — install-from-GUI** — fetch + place assets, confirmation dialog, restart-if-required handling. ✅ shipped in core (v0.51.644 / nesquena/hermes-webui#4746). `POST /api/extensions/install` downloads the registry zip, verifies sha256 against the registry entry, blocks zip-slip/symlink-escape/cross-host-redirect, extracts, and tracks installed files in `extension-install-manifest.json` for clean uninstall. The library delivery format (#9 — zip + artifact `sha256`) is what it consumes.
- [x] **Phase 4 — uninstall** + polish. ✅ shipped in core (v0.51.644 / nesquena/hermes-webui#4746). `POST /api/extensions/uninstall` removes exactly the tracked file set (bottom-up empty-dir cleanup), surfaced via the **Installed** tab. _(Later "auto-install of trusted" remains a future nicety, not part of this issue's scope.)_

## Notes / dependencies

- Depends on the **registry** (`registry.json`) existing — see the registry generator issue.
[...]

PR #10, opened 2026-06-21
RFC 0001: extension.json entry-metadata schema (for discussion)

nesquena-hermes:
## Summary

An **RFC (discussion, not implementation)** proposing a concrete `extension.json` per-extension metadata schema, so the team has something specific to react to rather than a blank page.

Kept as documentation under `docs/rfcs/0001-extension-json-schema.md`. Nothing is wired up — this is purely to align on the shape before we build the registry generator (#3), CI validation (#6), and safety gates (#8), all of which need this file.

## What it proposes

- A per-extension `extension.json` with identity, `assets`, `capabilities`, optional `sidecar`, `restart_required`, `screenshots`, and an honest `permissions` disclosure block.
- **Recommendation (open to debate):** authors write the one rich `extension.json`; the registry Action *derives* the minimal runtime `manifest.json` the core loader consumes. Single source of truth for authors; keeps the loader contract minimal. The two-hand-written-files alternative is explicitly on the table.
- The fields the Action *adds* to the published registry entry (`download`, `sha256`, etc.) vs. what authors write.
- A first cut at what CI / safety gates validate.

## Please weigh in on the open questions

1. One file (author-writes-`extension.json` + derive manifest) or two hand-written files?
2. Initial `capabilities` vocabulary + who owns the canonical list.
3. `permissions` granularity for the gallery trust display.
4. Versioning / "update available" mechanics.
5. Screenshots in-repo vs. referenced.

Relates to #3, #6, #8. Out of scope here: the proxy security model (#5) and install mechanics (#4/#9).

Not looking to merge this quickly — it's here to collect input. @franksong2702 / @Santastabber, your take especially on Q1 (you author against it) and Q2.

### DIFF
--- a/docs/rfcs/0001-extension-json-schema.md
+++ b/docs/rfcs/0001-extension-json-schema.md
@@ -50,9 +50,16 @@ the table.
   },
 
   // --- what core surface this entry needs (capability names, not version pins) ---
-  "capabilities": ["manifest-bundle", "sidecar"],   // e.g. manifest-bundle | sidecar | proxy
+  // capabilities are CORE-OWNED: the canonical list lives in the core repo, since
+  // it defines the actual surface. Only list a capability core has SHIPPED.
+  // `manifest-bundle` and `loopback-sidecar` (direct browser→loopback, shipped today)
+  // are valid now; do NOT list `sidecar-proxy` until core ships the same-origin proxy.
+  "capabilities": ["manifest-bundle", "loopback-sidecar"],
 
-  // --- optional local helper process (descriptive until the core sidecar/proxy contract ships) ---
+  // --- optional local helper process. The `sidecar` block is descriptive metadata. ---
+  // Today an extension reaches this directly (browser → loopback, allowed by the
+  // core CSP connect-src). A same-origin proxy to it is FUTURE work (capability
+  // `sidecar-proxy`), not assumable by entries yet.
   "sidecar": {
     "type": "loopback",
     "origin": "http://127.0.0.1:17787",
@@ -64,11 +71,18 @@ the table.
 
   // --- gallery + trust ---
   "screenshots": ["screenshots/pet.png"],
-  "permissions": {                      // honest disclosure; drives review + the gallery's "this extension can…" note
-    "network": false,
-    "filesystem": false,
-    "native_host": true,               // true here because of the sidecar
-    "sidecar": true
+  // Honest disclosure; drives review + the gallery's "this extension can…" note.
+  // Finer-grained than a coarse 4-bool set BECAUSE extension JS runs in the WebUI
+  // origin and can call authenticated WebUI APIs — that authority must be surfaced
+  // explicitly, not hidden inside a generic "network" bool. (per @santastabber)
+  "permissions": {
+    "webui_authenticated_api": true,   // calls authenticated same-origin WebUI APIs (the in-origin authority)
+    "webui_dom": ["composer", "sidebar"],  // which UI surfaces it reads/mutates (coarse named surfaces)
+    "network_external": false,         // any non-same-origin / non-loopback fetch
+    "loopback_sidecar": true,          // talks to a declared localhost sidecar
+    "filesystem": false,               // reads/writes local files (via a sidecar/native host)
+    "native_host": true,               // starts or drives a native/desktop process
+    "storage": ["localStorage"]        // browser storage it uses
   },
 
   // --- compatibility ---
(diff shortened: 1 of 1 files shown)


santastabber:
Useful RFC and I like the one-author-file direction, but I’d keep this clearly draft until we settle a few contracts:

- capabilities should be core-owned and should not include `proxy` until core ships that capability;
- distinguish `loopback-sidecar` from any future `sidecar-proxy`;
- permissions need more granularity than `{network, filesystem, native_host, sidecar}` because extension JS runs in the WebUI origin and can call authenticated APIs;
- the generated runtime manifest must preserve the current core hardening rules for asset paths and manifest safety;
- install metadata should account for artifact integrity, zip-slip-safe extraction, installed file tracking, rollback, and uninstall.

nesquena-hermes:
All five land — agreed on every point, and I've folded them into the RFC (pushed). Keeping it draft. Point by point:

1. **Capabilities core-owned + no `proxy` until shipped** — yes. Updated the schema so `capabilities` is explicitly core-owned (canonical list lives in the core repo since it defines the surface), and validation rejects any capability core hasn't shipped. The example now uses `manifest-bundle` + `loopback-sidecar`; `sidecar-proxy` is called out as future/gated.

2. **`loopback-sidecar` vs future `sidecar-proxy`** — done, and I verified the distinction against core: the CSP `connect-src` already allows `http://127.0.0.1:*`/`localhost`/`ws://` loopback, so **direct browser → loopback is the shipped pattern today** (`loopback-sidecar`), and the same-origin `/extensions/<id>/sidecar/*` proxy is a separate future capability (`sidecar-proxy`) entries can't assume yet. Mirrored this in VISION (PR #2) too.

3. **Permissions granularity** — agreed the coarse 4-bool set was too blunt precisely because extension JS runs in the WebUI origin and can call authenticated APIs. Replaced it with a finer block that surfaces that authority explicitly: `webui_authenticated_api`, `webui_dom` (named surfaces), `network_external`, `loopback_sidecar`, `filesystem`, `native_host`, `storage`. First cut — naming/buckets very much open for your input (Q3).

4. **Derived manifest preserves core hardening** — added an explicit section: the Action-derived runtime manifest must satisfy *every* rule the core loader enforces (same-origin paths, no traversal/encoded dot-segments, URL-count + size caps, bare-path resolution). Held to the same bar as a hand-written manifest; if an entry's assets can't produce a core-valid manifest, it fails validation.

5. **Install metadata** — added a lifecycle section covering artifact integrity (`sha256` verified before extract), zip-slip-safe extraction (confined to the extension dir, reject `..`/absolute members), installed-file tracking, and rollback + clean uninstall. The mechanics live in the delivery issue (#9) and Settings-UI issue (#4); the schema/RFC now names what they must support.

Still genuinely open and would value your call (you author against this): **Q1 one-file-vs-two** (author `extension.json` + derive manifest, vs. hand-write both), and the exact **permissions vocabulary** in #3.

franksong2702:
I checked this against the current Desktop Companion code rather than answering abstractly.

I still prefer the one-file model: authors maintain `extension.json`, and CI derives the runtime loader manifest. For Desktop Companion, the author-facing metadata is much richer than the loader manifest: identity, assets, sidecar, permissions, compatibility, source repo, screenshots, install notes, and lifecycle notes all need to stay together. Hand-writing both `extension.json` and `manifest.json` would be easy to drift.

For current capabilities, Desktop Companion would declare only:

- `manifest-bundle`
- `loopback-sidecar`

It should not declare `sidecar-proxy` until core ships that capability.

On permissions, the current finer-grained direction is right, but Desktop Companion shows why a few fields may need to be more specific than booleans:

- It calls authenticated WebUI API today via `/api/sessions`.
- It creates extension-owned DOM only; it does not replace core views.
- It can navigate/select a session via `window.loadSession(sessionId)` when available, with a URL/localStorage fallback.
- It uses localStorage for companion UI preferences, and currently also touches the existing `hermes-webui-session` key for session navigation fallback.
- It talks to a loopback sidecar at `http://127.0.0.1:17787`.
- The sidecar serves bundled/static pet assets and keeps current state in memory; it does not write arbitrary local files.
- The desktop pet host is a real native host/Tauri app: transparent always-on-top windows, menu handling, drag behavior, and process restart.
- No external network access is needed for the current shape.

So I would like the permissions vocabulary to eventually distinguish at least:

- WebUI API access by purpose, e.g. session/status read vs chat/action write
- extension-owned DOM vs mutation/replacement of core views
- extension-owned storage vs touching existing WebUI storage keys
- loopback sidecar access
- native host/windowing behavior
- arbitrary filesystem access vs serving bundled extension assets

One lifecycle note from Desktop Companion: `restart_required` may need to be precise about what restarts. Installing/enabling WebUI assets, starting a loopback sidecar, launching a native host, and remembering that the native pet was intentionally closed are separate states. It may be useful to distinguish WebUI restart required from sidecar/native-host start required, and to leave native-host autostart preference owned by the extension rather than WebUI core.

For screenshots, in-repo is fine for the first pass. For updates, I’d use `version` for the user-facing comparison and `sha256` for integrity. Compatibility should stay capability-first, with `min_webui_version` only as a fallback hint.

nesquena-hermes:
This is great — grounding it in the actual Desktop Companion surface is exactly what the RFC needed. Folded it all in (pushed):

**One-file model — settled.** Both you and @santastabber land on author-maintains-`extension.json`, CI derives the loader manifest. Moved it from open-question to the "Resolved (maintainer consensus)" section, with your drift rationale (the author-facing metadata is much richer than the loader manifest, so two hand-written files would drift).

**Capabilities** — `manifest-bundle` + `loopback-sidecar` only, no `sidecar-proxy` until core ships it. Already gated that way in validation; your confirmation locks it.

**Permissions — reworked to be purpose-based**, using your Desktop Companion breakdown as the concrete model:
- `webui_api: { read: [...], write: [...] }` — session/status *read* vs chat/action *write*, distinct buckets
- `webui_navigation` — `window.loadSession` etc., separate from API writes
- `dom: { owned, mutates_core_views }` — extension-owned DOM vs replacing core views
- `storage: { owned: [...], shared_webui_keys: [...] }` — so touching `hermes-webui-session` for nav-fallback is disclosed distinctly from your own `companion-prefs`
- `loopback_sidecar`, `native_host` (windowing/menus/drag/restart), and `filesystem: { arbitrary, serves_bundled_assets }` so "serves bundled pet assets" reads differently from arbitrary local file access
- `network_external: false`

**Lifecycle — split per your note.** Replaced the single `restart_required` with a `lifecycle` block: `webui_restart_required` / `sidecar_start_required` / `native_host_start_required` as separate states, plus `native_host_autostart: "extension_owned"` to keep the "pet was intentionally closed" preference owned by the extension, not WebUI core. That felt like the sharpest insight — the install flow shouldn't assume "install = one restart."

**Screenshots in-repo first pass, `version` for update comparison + `sha256` for integrity, capability-first compat** — all adopted as resolved.

Remaining open (narrowed): the final permissions naming, whether the `lifecycle` split matches how the install flow (#4) should reason about start/restart states, and where older artifact versions live. Your Desktop Companion entry will be the real test of whether this vocabulary actually fits — looking forward to it being the first entry that exercises it end to end.

Issue #7, opened 2026-06-21
First real entry: Desktop Companion (sidecar-class extension, to spec)


nesquena-hermes:
## Goal

Land **Desktop Companion** (the digital-pet assistant) as the first real, fully-to-spec extension entry in this repo. It's the canonical *richer / sidecar-class* test case and the agreed definition of success for the whole system.

> Success criterion (Nathan): "We'll know we've succeeded when Frank can submit the pet/Desktop Companion extension fully to spec into the extensions repo, and Santastabber can submit a few of his too."

## Shape

- Manifest-bundled WebUI assets (the in-origin frontend) + a **trusted local loopback sidecar** (Tauri shell / local process) for the desktop surface.
- Currently standalone at `franksong2702/hermes-webui-desktop-companion`; intended to be adapted to whatever manifest/registry/sidecar conventions this repo settles on.

## What it exercises (and validates) end-to-end

- [ ] manifest-bundled assets (✅ contract exists)
- [ ] `extension.json` registry metadata (depends on the registry schema issue)
- [ ] sidecar metadata declaration (`type: loopback`, `origin`, `health_path`)
- [ ] the sidecar proxy path, once Tier A lands (depends on the proxy issue)
- [ ] install / disable / uninstall docs + `restart_required` behavior
- [ ] clear trust-model note (what it touches, that it runs a local process)

## Dependencies / sequencing

- Can land as a **documented entry + assets** as soon as the registry schema is set, even before the proxy exists.
- The sidecar/proxy integration becomes real once the backend-proxy issue (Tier A) ships. Treat Desktop Companion as the driving use case that keeps that work honest.

## Companion: Santastabber's entries

Track the UI-tweak-class entries alongside as the second validation set — Templates (wants backend routes), Codex/Claude usage sidebar meters, paperclip + git sidebar items. (The mobile floating-conversation button is also proposed for core as nesquena/hermes-webui#4509; decide per-item whether it's core or extension.)

Part of the extensions roadmap (see VISION.md §success-criteria).

nesquena-hermes:
**Progress update (2026-06-22)**

Desktop Companion is well advanced as the first sidecar-class entry:

- **Standalone source implementation merged** — `franksong2702/hermes-webui-desktop-companion#3` ("Stabilize Desktop Companion bridge", merged `e9861f0b`). Sidecar + native host live there (not vendored into this library, by design).
- **Extension entry PR #11 is in active review.** It's been through several iterations:
[...]

Issue #9, opened 2026-06-21
Install delivery: how one-click install fetches + verifies extension assets (zip + sha256 recommended)


nesquena-hermes:
## Goal

Decide and implement how a user's one-click **Install** actually gets an extension's files onto their machine. This is the step that turns "it's in the curated repo" into "it's installed and running in my WebUI."

## Recommended approach (for discussion)

**Published artifact + integrity hash — not clone, not submodule.**

- The registry GitHub Action (deploying to GitHub Pages) also publishes a **per-extension artifact** (zip of the extension's assets) alongside `registry.json`.
- Each registry entry carries a **`download` URL + `sha256`**.
- Install flow: fetch the artifact → **verify the `sha256` against the vetted registry entry** → extract into the local extension dir (`HERMES_WEBUI_EXTENSION_DIR`) → enable via the manifest → restart if `restart_required`.

The hash is the key safety property: it binds the downloaded bytes to the entry that passed review, so even a Pages/CDN compromise can't deliver unreviewed code that installs cleanly.

Rejected alternatives:
- **Clone the whole repo** — heavy, pulls everything, no per-extension isolation.
- **Git submodule** — too much ceremony for an end-user install; wrong tool.

## Open questions (maintainer input wanted)

- Zip-per-extension (recommended) vs. per-file fetch from the Pages site?
- Where do artifacts live — Pages alongside the registry, or GitHub release assets?
- How does uninstall reverse it cleanly (track installed file set per extension)?
- Versioning/update path — how does the GUI offer "update available" when a registry entry's version/hash changes?

## Notes

- Core-side install mechanics live in `nesquena/hermes-webui` (the Settings → Extensions flow); this issue tracks the delivery contract the registry must provide.
- Depends on the registry generator + the integrity hash from the safety-gates issue.

Part of the extensions roadmap (see VISION.md). Pairs with the registry generator, Settings UI, and safety-gates issues.

nesquena-hermes:
Tracked in the extension-system roadmap umbrella: #12.

nesquena-hermes:
The core install-client half (fetch → verify sha256 → zip-slip-safe extract → installed-file tracking → rollback/uninstall) is tracked in nesquena/hermes-webui#4746; this issue owns the library-side delivery format (per-extension zip + sha256 in the registry).

nesquena-hermes:
## Status — this is now the keystone of the remaining chain

[...]

Issue #3, opened 2026-06-21
Registry: per-extension metadata schema + generated registry.json (GitHub Action)


nesquena-hermes:
## Goal

Define the per-extension metadata schema and a GitHub Action that generates a single `registry.json` index on every merge to `main`. This is the foundation the in-WebUI gallery + install flow reads from.

## Design (from VISION.md §registry)

- **Per-extension metadata file** — each `extensions/<id>/` carries an `extension.json` registry entry. Proposed fields:
  - `id`, `name`, `description`, `version`, `author`, `homepage`
  - `capabilities` — required extension API surfaces (e.g. `manifest-bundle`, `sidecar`, `proxy`) — capability names preferred over version pins
  - `assets` — scripts/stylesheets (same shape the loader reads)
  - `sidecar` (optional) — `{ type, origin, health_path }`
  - `restart_required` (bool) — drives the install flow's restart step
  - `screenshots`, and trust disclosures (network / filesystem / native-host / sidecar)
  - (Open: whether `extension.json` and the loader's `manifest.json` stay separate — keeps the loader contract minimal — or merge.)
- **Generator Action** — on push to `main`: discover every `extensions/*/extension.json`, validate each against the JSON schema (fails the build on invalid), emit top-level `registry.json`.
- **Publish location** — recommend GitHub Pages or a pinned release asset so WebUI fetches a stable, versioned/cacheable URL rather than a mutable raw-`main` file. (Needs sign-off.)
- **Trust boundary = merge** — the registry only ever contains merged (vetted) entries, so the install flow can trust the registry wholesale; no separate "trusted" flag.

## Tasks

- [ ] Draft the `extension.json` JSON schema
- [ ] Decide `extension.json` vs `manifest.json` separation (or merge)
- [ ] Write the registry generator script
- [ ] Wire the generator into a GitHub Action on push to `main`
- [ ] Decide + implement the publish location for `registry.json`
- [ ] Document the schema in `docs/`

## Open questions

- Exact per-entry schema fields (above is a proposal).
- Publish location: committed back to `main` vs. GitHub Pages vs. release asset.

Part of the extensions roadmap (see VISION.md). Pairs with the CI-validation issue and the Settings → Extensions UI issue.

nesquena-hermes:
Decisions from Nathan (2026-06-21) narrowing this issue:

[...]

PR #11, opened 2026-06-22
[extension] Add Desktop Companion entry

franksong2702:
## Summary

- add `extensions/desktop-companion/` as the first sidecar-class extension entry
- include the WebUI bridge script, pet sprite assets for the native host, runtime `manifest.json`, and PR #10-style `extension.json` metadata
- document install, disable, uninstall, trust model, sidecar health, lifecycle, permissions, compatibility, and verification notes
- update repository docs now that a real extension entry exists
- sync the bridge adapter from `franksong2702/hermes-webui-desktop-companion`

## Scope

This is a documented entry + WebUI bridge script. It intentionally does not vendor the Tauri/native host source tree; the sidecar and native host remain in `franksong2702/hermes-webui-desktop-companion`.

The WebUI extension is intentionally invisible: it does not render a browser pet, overlay, attention cards, or browser UI. It only reads guarded WebUI session state and posts snapshots to the local sidecar so the native desktop pet can react on the desktop.

The user-visible desktop pet experience requires the loopback sidecar and native desktop pet host from the source repo. The WebUI adapter alone is only the browser bridge.

It declares only currently shipped capabilities:

- `manifest-bundle`
- `loopback-sidecar`

It does not declare `sidecar-proxy` because Hermes WebUI core has not shipped that capability.

Relates to #7 and aligns with the current `extension.json` shape in #10.

Source implementation PR: https://github.com/franksong2702/hermes-webui-desktop-companion/pull/3

## Validation

From this repo:

- `node scripts/validate-desktop-companion.mjs`
- `python3 -m json.tool extensions/desktop-companion/extension.json`
- `python3 -m json.tool extensions/desktop-companion/manifest.json`
- `node --check extensions/desktop-companion/assets/companion-adapter.js`
- `git diff --check`

From the Desktop Companion source repo:

- `npm test` passes 26/26
- `node --check extension/companion-adapter.js`
- `node --check desktop-pet/web/bubbles.js`
- `node --check desktop-pet/web/pet.js`
- `python3 -m json.tool extension/extension.json`
- `python3 -m json.tool extension/manifest.json`
- `git diff --check`
- `npm run desktop:build`
- Claude Code review: no blocking issues found after the Tauri-webview/browser-overlay architecture clarification

Manual testing covered desktop-pet running, ready, approval, clarify, quick reply, browser-tab focus, and no browser in-page pet/overlay behavior during the local 8787 extension-mode test pass.

## Notes

The extension can run without the sidecar and fails closed invisibly. Desktop behavior requires starting the local sidecar and native host from the companion source repo. One-click install, settings lifecycle management, and sidecar proxy support remain future WebUI/core work.


nesquena-hermes:
## Deep review — Desktop Companion entry (the first sidecar-class extension) 🐾

Thanks for this, Frank — this is the entry the whole library has been building toward (the §8 success criterion: "Frank can submit the pet/Desktop Companion extension fully to spec"). I reviewed it as **trusted application code** (extension JS runs in the WebUI origin against the authenticated session), against the repo conventions (`README` / `CONTRIBUTING` / `docs/extension-entry.md`), the RFC 0001 `extension.json` schema (#10), and the real core loader (`api/extensions.py`) — running the manifest through the live loader rather than trusting the prose. Verifications I ran are noted inline.

**Verdict: strong, and very close. One real (forward-looking) portability bug worth fixing now, plus a few small polish items. Holding for those + a maintainer's call, not for a rewrite.**

---

### What's excellent

- **`extension.json` tracks RFC 0001 almost exactly** — identity, `assets`, `capabilities`, `sidecar`, the split `lifecycle` (`webui_restart_required` / `sidecar_start_required` / `native_host_start_required` / `native_host_autostart: extension_owned`), and the fine-grained `permissions` block. This is the schema's first real-world exercise and it fits.
- **Capability honesty.** Declares only `manifest-bundle` + `loopback-sidecar` (both shipped); correctly does **not** declare `sidecar-proxy`. README and PR body both call this out. Exactly right per the core-owned-capabilities rule.
- **The `permissions` block is honest against the code.** I cross-checked each claim against `companion-adapter.js`: `webui_api.read:["sessions"]` ↔ `fetch('/api/sessions')` only, `write:[]` ↔ no mutating calls; `dom.owned:true / mutates_core_views:false` ↔ it only appends its own `<aside id="hermes-webui-desktop-companion">`; `storage.owned` / `shared_webui_keys` ↔ the exact keys read/written; `network_external:false` ↔ the only non-same-origin fetch is the declared loopback. No undisclosed surface. This is the disclosure bar other entries should copy.
- **XSS hygiene is correct.** `renderCards()` builds `innerHTML`, but every interpolated value (`session_id`, `status`, `title`, `text`) goes through `escapeHtml()`. Session titles/text are attacker-influenceable (they come from session content), so this matters — and it's handled.
- **Defensive throughout:** IIFE + `__HERMES_WEBUI_DESKTOP_COMPANION_LOADED__` double-load guard; every `localStorage` access wrapped in try/catch; `fetch('/api/sessions')` failure fails closed (empty attention, pet shows idle/offline) rather than throwing; loopback POST uses `credentials:'omit'` so the authed session cookie is never sent cross-origin to the sidecar.
- **Sidecar/native host correctly NOT vendored** — kept in `franksong2702/hermes-webui-desktop-companion`, only the WebUI-facing assets live here. Matches the scope the PR body states and the trust model (no unreviewed binaries in the library).
- **I verified the spritesheets are real binaries** (courier 657 KB, keeper 1.08 MB, shiba 2.39 MB), not 0-byte placeholders — the pet actually renders.
- **The CSP dependency the README claims is real.** I grepped core `api/helpers.py`: `_CSP_CONNECT_BASE` includes `http://127.0.0.1:* http://localhost:*`, so the `POST http://127.0.0.1:17787/api/webui/snapshot` works today without any core change. Good that the README lists "loopback CSP allowance" as a required surface.

---

### 🐛 The one real finding — hardcoded sprite paths break in a multi-extension layout

`companion-adapter.js` hardcodes the pet sprites at the **root** of `/extensions/`:

```js
spritesheetUrl: '/extensions/pets/keeper/spritesheet.webp'   // (and shiba, courier)
```

I ran this through the live loader both ways:

- **Documented single-extension install** (`HERMES_WEBUI_EXTENSION_DIR=.../desktop-companion`): the loader serves `/extensions/pets/...` from `<DIR>/pets/...` → **resolves correctly. Works today.** ✅
- **Multi-extension bundle layout** (the repo's actual end-state — `DIR` points at a parent holding many entries, assets under `/extensions/desktop-companion/...`): the loader resolves the *script/style* paths relative to the manifest entry, so they correctly become `/extensions/desktop-companion/assets/companion-adapter.js`. But the **hardcoded sprite URLs stay `/extensions/pets/...`** → the loader looks under `<DIR>/pets/...`, the files are actually at `<DIR>/desktop-companion/pets/...` → **the sprites 404 and the pet renders blank.** ❌ (verified against `api.extensions.get_extension_config()` + the static-serve path resolution.)

This isn't a blocker for the manual install the README documents, but it's a latent portability bug that bites the moment Settings → Extensions install (#4) lands — that flow almost certainly installs each entry into a per-id subdir under a shared root, which is exactly the layout that breaks. Two clean fixes:

1. **Derive the asset base from the adapter's own URL** (most robust — survives either layout):
   ```js
   const ASSET_BASE = new URL('.', document.currentScript ? document.currentScript.src : location.href).href;
   // …then resolve sprites relative to ASSET_BASE instead of a hardcoded '/extensions/pets/…'
   ```
   (Capture `document.currentScript` at top-of-IIFE evaluation, before any async, since it's null later. The sprites sit under `pets/` as a sibling of `assets/`, so resolve against `ASSET_BASE/../pets/<skin>/spritesheet.webp`.)
2. Or, lower-effort: make the base **configurable** via the existing `window.HERMES_DESKTOP_COMPANION_CONFIG` override and **document the assumed layout** in the README so the install flow can set it.

I'd lean (1) — it makes the entry layout-agnostic and is the pattern the rest of the library should adopt.

---

### Minor / polish (non-blocking)

- **`manifest.json` carries a `description` field** on the extension object. Harmless (the core loader's `_iter_manifest_entries` reads only `scripts`/`stylesheets`/`enabled` and silently ignores the rest), and it matches the RFC's "derive the runtime manifest from `extension.json`" direction — but worth noting it's inert today. Once the registry generator (#3) derives `manifest.json` from `extension.json`, this hand-written manifest becomes a generated artifact; fine to keep for now as the manual-install convenience.
- **`extension.json` `screenshots: []` is empty.** RFC 0001 lists `screenshots` as gallery metadata and the resolved decision was "screenshots in-repo for the first pass." For *the* flagship sidecar entry that will headline the future gallery, one committed `screenshots/pet.png` would be worth adding (it's also the natural place to show the three skins). Not a blocker.
- **Repo-size heads-up (for the maintainers, not a change request):** the three `.webp` sheets total ~4.1 MB committed to the library. RFC 0001 explicitly flagged "revisit if repo size becomes an issue" — Desktop Companion is the first data point on that, and it's a chunky one. Worth a maintainer note on whether large binary sprite assets stay in-repo or move to the per-extension zip/release-asset delivery (#9) once that exists.
- **`buildSnapshot()` data egress (disclosure completeness, not a bug):** the snapshot POSTed to the loopback sidecar includes `page.href`, `page.title`, and `companion.attention[]` (session titles + status text). It's localhost-only, `credentials:'omit'`, and the user runs the sidecar themselves — entirely consistent with the trust model — but since the entry sets the disclosure bar for the library, consider one README line under "Trust And Permissions" noting that session titles/status snapshots are sent to the local sidecar. (Permissions block already implies it via `loopback_sidecar:true`; this would just make the *what* explicit.)
- **`openSession()` URL fallback** sets `?session_id=` and writes `hermes-webui-session` — both already disclosed in `shared_webui_keys`. Good. Minor: the primary path `window.loadSession(sessionId)` depends on a global that core doesn't formally guarantee as an extension API; it's guarded (`typeof … === 'function'` + try/catch with URL fallback), so it degrades fine — just flagging that it's an undocumented core global, the kind of thing the "required extension API surface" compatibility section exists to track.

---

### Process / scope

- Lines up with #7 (first real entry) and is consistent with the #10 schema — I'd suggest cross-linking this PR on #7. It's effectively the empirical validation of RFC 0001, so any field that felt awkward to fill here is useful signal back on #10.
- This is a **convention-relevant** PR (first entry → it sets the de-facto template every future entry copies), so per our process I'm **not merging it unilaterally** — flagging for a maintainer's call (@nesquena / @santastabber). My recommendation: **fix the sprite-path portability issue (the one real bug), then it's good to merge**; the screenshot + the two README disclosure lines are nice-to-haves that could land in the same push or a fast follow-up.

Really nice work — the disclosure discipline and capability honesty here are exactly the standard that makes the "vetted == merged" model safe. The only thing standing between this and the reference entry for the whole library is making the sprite paths layout-agnostic.

franksong2702:
Thanks for the deep review. I pushed an update addressing the blocking portability issue.

Changes in the latest push:

- the adapter now derives pet sprite URLs from `document.currentScript.src` instead of hardcoding `/extensions/pets/...`
- it handles both current single-extension manual layout and future nested `desktop-companion/assets/...` layout
- added `scripts/validate-desktop-companion.mjs` to check manifest/metadata references, capabilities, lifecycle, pet assets, and both sprite URL layouts
- added the README disclosure that session attention summaries, page URL/title, and companion state are sent to the local sidecar

Validation rerun:

- `node scripts/validate-desktop-companion.mjs`
- JSON validation for `extension.json`, `manifest.json`, and pet manifests
- `node --check extensions/desktop-companion/assets/companion-adapter.js`
- Desktop Companion source repo `npm test` now passes 13/13 with the same adapter fix

I left screenshots as a follow-up so this stays focused on the merge-blocking portability fix.

nesquena-hermes:
## Re-review of the update (commit `136e644`) — portability fix verified ✅

Thanks for the fast turnaround, Frank. I re-reviewed the pushed changes empirically (not just reading the diff) and the blocking issue is genuinely resolved. Details of what I checked:

**The sprite-path fix is correct — verified in both layouts.** I replicated the new `ASSET_BASE` / `PETS_BASE` logic in Node against the two real `currentScript.src` values:

| Layout | Resolved sprite URL | Correct? |
|---|---|---|
| Single-extension (`DIR=…/desktop-companion`), script at `/extensions/assets/companion-adapter.js` | `/extensions/pets/keeper/spritesheet.webp` | ✅ matches where the loader serves it |
| Multi-extension bundle, script at `/extensions/desktop-companion/assets/companion-adapter.js` | `/extensions/desktop-companion/pets/keeper/spritesheet.webp` | ✅ the exact path that 404'd before |
| `currentScript` null (defensive fallback) | `/extensions/pets/keeper/spritesheet.webp` | ✅ sensible default |

**And I confirmed the fix is robust against how core actually injects the adapter** (this is the thing that could have quietly broken a `currentScript`-based approach): core (`api/extensions.py:917-923`) writes a **static `<script src="…" defer></script>`** into the page body — not a dynamically-`appendChild`'d or `async`/module script. For a static *deferred classic* script, `document.currentScript` is reliably populated during execution (defer changes timing, not `currentScript`), so `CURRENT_SCRIPT_SRC` will be set in the real load path. Good instinct capturing it synchronously at the top of the IIFE.

**Other items confirmed landed:**
- `node scripts/validate-desktop-companion.mjs` → passes (I ran it on a clean checkout of the branch — exit 0). Nice touch asserting `doesNotMatch(/\/extensions\/pets\//)` so a regression to hardcoded paths fails CI, and pinning both layout resolutions. That's effectively a down payment on #6.
- README disclosure (lines 101–102) now explicitly states the snapshot sends page URL/title, companion state, and session attention summaries incl. titles + status text. Exactly the disclosure-completeness point. ✅
- `extension.json` / `manifest.json` still valid JSON; capabilities unchanged (`manifest-bundle` + `loopback-sidecar`, no `sidecar-proxy`); no leftover hardcoded `/extensions/pets/` anywhere in the adapter.

---

### One small robustness nit on the fix itself (optional, non-blocking)

The layout branch keys off the substring `/assets/` in the script path:

```js
const PETS_BASE = CURRENT_SCRIPT_SRC && new URL(CURRENT_SCRIPT_SRC).pathname.includes('/assets/')
  ? new URL('../pets/', ASSET_BASE)
  : new URL('pets/', ASSET_BASE);
```

This is correct for both shipped layouts. The only way it misfires is if a future install root itself contained `/assets/` as an ancestor segment (e.g. `…/assets/desktop-companion/assets/companion-adapter.js`) — `includes()` matches the first occurrence, so the `../pets/` hop could land one level off. Extremely unlikely given the repo convention, so I would **not** block on it, but if you want to make it bulletproof: anchor the check to the adapter's own directory rather than a substring — e.g. test whether the script's immediate parent dir is `assets` (`pathname.replace(/\/[^/]*$/,'').endsWith('/assets')`), which only matches the real `assets/` that holds the adapter. Your call — current code is fine for everything we actually ship.

---

### Net

The merge-blocking portability bug is fixed and verified, and the snapshot disclosure + a layout-pinning CI check both landed. From my side there's **nothing further I'd require before merge** — the only remaining items are the ones already agreed as follow-ups (a `screenshots/pet.png` for the gallery) and the maintainer-level note (whether the ~4 MB of in-repo sprite binaries stay here vs. move to zip/release delivery once #9 exists), neither of which should hold this PR.

Still flagging for a maintainer's merge call rather than merging myself, since this is the convention-setting first entry — but my recommendation is now **ready to merge as-is** (the `/assets/` nit above is genuinely optional). Great work; this is the reference entry the rest of the library can copy.

franksong2702:
I pushed one more correction after re-checking the product intent: Desktop Companion is meant to be a desktop pet, not a browser pet.

The reason for this correction is that some Browser Pet code from the early development phase was carried over during the migration into the standalone Desktop Companion / extension-library path. That was an implementation carryover, not the intended product shape. The product intention is Desktop Pet: WebUI provides state, the sidecar carries that state locally, and the native desktop pet is the user-visible surface.

The latest push removes the in-page pet/overlay from the WebUI extension entirely:

- no browser DOM is created
- no WebUI CSS asset is loaded
- no browser attention cards, badge, skin switcher, or `window.loadSession` navigation
- `extension.json` now declares `dom.owned:false`, `webui_navigation:false`, and no extension-owned storage keys
- the WebUI script is bridge-only: it reads `/api/sessions`, derives attention state, and POSTs snapshots to the local sidecar
- native desktop pet rendering remains in the standalone Desktop Companion source repo

Validation rerun:

- Desktop Companion source repo: `npm test` passes 13/13
- `node scripts/validate-desktop-companion.mjs`
- JSON validation for entry metadata/manifest/pet manifests
- `node --check extensions/desktop-companion/assets/companion-adapter.js`

This is a product-scope correction, not just polish: the extension entry now matches the intended desktop-pet architecture.

nesquena-hermes:
## Final review of the latest push (commit `ce49749`) — approve ✅

Took a final pass over the latest version. This push is the **browser-pet → desktop-pet scope correction** (the WebUI side is now a bridge-only adapter, with the pet rendering moved entirely to the native host). I re-reviewed it as application code rather than trusting the description — verifications inline.

**This is a genuine improvement, not just a trim.** Removing the in-page rendering shrinks the extension's WebUI footprint and trust surface substantially, and the metadata + docs were brought into lockstep with it.

**What I verified in the new code:**

- **Bridge-only, confirmed.** `companion-adapter.js` (449 → 199 lines) now only: reads `/api/sessions` (`credentials:'include'`), derives attention state, and POSTs a snapshot to the loopback sidecar (`credentials:'omit'`). No `createElement` / `appendChild` / `innerHTML` anywhere — **no DOM is created**, and the CSS asset is gone (`stylesheets: []`, file removed from the PR). ✅
- **The whole XSS surface is moot now** — the previous `renderCards()` / `escapeHtml()` HTML-rendering path no longer exists, so there's nothing to escape. Strictly fewer ways to go wrong.
- **The sprite-path portability bug I flagged earlier is resolved by removal** — the WebUI adapter no longer resolves or loads pet sprites at all (the `.webp` sheets stay in-repo for the native host to consume), so there's no `/extensions/pets/...` path to get wrong. The `/assets/` substring nit from my last comment is also moot.
- **No navigation / no extension-owned storage** — `window.loadSession`, the URL/`hermes-webui-session` fallback, skin switcher, and badge are all gone. The adapter only *reads* `hermes-session-viewed-counts` and `hermes-session-completion-unread`; there's **no `localStorage.setItem` anywhere**. Status is now surfaced via a read-only `window.__HERMES_WEBUI_DESKTOP_COMPANION_STATUS__` global instead of DOM — clean.
- **`extension.json` permissions match the slimmed code exactly** (cross-checked field by field): `dom.owned:false`, `webui_navigation:false`, `storage.owned:[]` with `shared_webui_keys` correctly narrowed to the two keys it actually reads (and `hermes-webui-session` correctly dropped since the nav-fallback is gone). `webui_api.read:["sessions"]`, `write:[]`, `network_external:false`, `loopback_sidecar:true`, `native_host:true`, capabilities `manifest-bundle`+`loopback-sidecar` (still no `sidecar-proxy`) — all consistent. The disclosure honesty stays exemplary.
- **Docs in lockstep.** README now opens with "It does not render a browser pet or WebUI overlay," the architecture diagram and verification steps reflect bridge-only, and I found no leftover references to the removed skin/card/badge/navigation pieces.
- **Validation green on a clean checkout:** `node scripts/validate-desktop-companion.mjs` → passes (exit 0); all JSON (entry/manifest/3 pet manifests) valid; `node --check` clean. Frank's note that the source repo's `npm test` passes 13/13 with this adapter.

**Net:** I approve of these changes. The scope correction makes the entry simpler, lower-risk, and a cleaner reference for the library — and it stays the canonical sidecar-class example. Nothing further from me before merge.

Per process I'm **not merging** this myself (convention-setting first entry → maintainer's call, @nesquena / @santastabber). The only outstanding items remain the already-agreed follow-ups (a `screenshots/` asset for the future gallery; the maintainer-level note on whether the ~4 MB of in-repo sprite binaries stay here or move to zip/release delivery once #9 lands) — neither blocks this. Recommendation stands: **ready to merge as-is.** Nice correction, Frank.

franksong2702:
Source implementation has landed in the standalone Desktop Companion repo: https://github.com/franksong2702/hermes-webui-desktop-companion/pull/3

I have updated this extension entry to match that implementation and marked the PR ready for review.

A few scope notes for reviewers:

- the WebUI asset is bridge-only and does not render an in-page pet or browser overlay
- user-visible desktop behavior requires the local loopback sidecar plus the native desktop pet host from the source repo
- this entry declares only currently shipped capabilities: `manifest-bundle` and `loopback-sidecar`
- it does not declare `sidecar-proxy`, one-click install, or WebUI-managed sidecar lifecycle yet

Latest validation is listed in the PR body, including source repo tests/build and the extension-entry checks.

nesquena-hermes:
## Re-review of the latest push (commit `d8bc2bc`) — scope has materially expanded; metadata needs to catch up before merge

Thanks for the update, Frank — and congrats on landing the standalone source PR (franksong2702/hermes-webui-desktop-companion#3, merged `e9861f0b`). Flagging up front: my earlier "ready to merge as-is" was against the **199-line read-only bridge** (commit `ce49749`). This push takes the adapter to **~1,130 lines (+979/-45)** and changes what the extension *does* in a way that moves the trust posture, so it needs a fresh look rather than carrying the prior approval forward. I re-reviewed the actual diff.

### What's good (still holds)

- **No in-page DOM / pet came back** — 0 `createElement`/`appendChild`/`innerHTML`; it's still "no browser overlay." ✅
- **No `eval`/`Function`/dynamic-script, no external origins** — every non-same-origin call is the declared `127.0.0.1:17787` loopback, sidecar calls use `credentials:'omit'`, WebUI calls use `credentials:'include'`. ✅
- `node --check` clean; the standalone repo's tests/build are green per your note.

### The substantive change: this is no longer a read-only bridge — it's bidirectional control

The new adapter polls the loopback sidecar for **commands and executes them against the authenticated WebUI session**:

- `GET /api/pet/navigation` → `applyPetNavigationCommand()` → can `window.loadSession(sid)`, inject composer draft text, and — when `command.autosend` is set — **call `send()`, i.e. actually send a chat message as the user** (lines ~806-827, 846-867).
- `GET /api/pet/actions` → `executePetActionCommand()` → **POSTs `/api/approval/respond` and `/api/clarify/respond`** (lines ~941-967), i.e. the pet can answer approval prompts and clarify dialogs on the user's behalf.
- Plus `POST /api/session/draft` writes, and it ACKs each command back to the sidecar.

That's a coherent, non-malicious feature and the loopback-only + `credentials:'omit'`-to-sidecar boundary is the right shape. But it's a real escalation from "reads session state and posts read-only snapshots" to "the local pet can drive the session — switch sessions, send messages, and respond to approval/clarify gates." It deserves to be called out and decided on explicitly, not slipped in under a metadata block that says the opposite.

### 🔴 Blocking: `extension.json` permissions now contradict the code

The whole "vetted == merged" model rests on the structured disclosure being honest, and this is the entry that sets the template. Right now the metadata still describes the old read-only bridge:

| `extension.json` says | The code at `d8bc2bc` actually does |
|---|---|
| `webui_api.write: []` | POSTs `/api/session/draft`, `/api/approval/respond`, `/api/clarify/respond` |
| `webui_navigation: false` | `window.loadSession()`, `switchPanel('chat')`, `window.location.assign()` |
| `storage.owned: []` | `localStorage.setItem` for `hermes-pet-navigation-last-id` + `hermes-pet-action-last-id` |
| (nothing about inbound commands) | polls the sidecar for nav+action commands and executes them, incl. `autosend` and approval/clarify responses |

Please update the entry so the disclosure matches reality — concretely:
- `webui_api.write`: list `session_draft`, `approval_respond`, `clarify_respond` (whatever names we settle on)
- `webui_navigation: true`
- `storage.owned`: the two `hermes-pet-*-last-id` keys
- and surface the **inbound-command / can-drive-session** capability explicitly. The snapshot's own `capabilities.canReceiveActions` is the honest signal here — it should read `true` now, and the permissions block needs a field that conveys "the sidecar can issue navigation/action commands this adapter executes against the authed session." This is exactly the kind of authority RFC 0001 §permissions said must be surfaced *because* extension JS runs in the WebUI origin against the authenticated session.

### Other items

- **README** has one line that hints at it ("executes queued desktop-pet actions") but the "Current disclosed behavior" bullet list still reads as read-only. It should spell out: can switch sessions, inject + optionally auto-send composer drafts, and respond to approval/clarify prompts, all driven by the local sidecar. Honest, prominent disclosure on the flagship entry sets the bar for the library.
- **`window.send()` / autosend** is the highest-authority path here (it sends a message as the user). Worth a sentence in the README on the expected guardrails — e.g. that autosend is sidecar-initiated and the sidecar is a localhost process the user runs. Not asking to remove it; asking to disclose it clearly.
- **Validator (`scripts/validate-desktop-companion.mjs`)** — now that there's a write/command surface, consider extending it to assert the `extension.json` permissions stay consistent with the adapter (e.g. if the adapter references `/api/approval/respond`, `webui_api.write` must be non-empty). That keeps this drift from recurring and is a down payment on the #6/#8 safety-gate idea (permissions cross-checked against a static scan).

### Process / recommendation

This is now a **scope decision, not just polish** — it's the first entry to exercise the "extension drives the authenticated session" surface, which is adjacent to the Tier-A proxy / backend-routes discussion (#5) and the safety-gates plan (#8). So per our process I'm **not merging**, and I'd specifically want @AvidFuturist's read on whether "sidecar can navigate + send messages + answer approval/clarify prompts" is in-scope for the first shipped entry, or something to gate/phase. My recommendation:

1. Fix the `extension.json` permissions + README disclosure to match the code (blocking — the metadata must not under-state the authority).
2. Nathan to confirm the bidirectional-control scope is what we want as the reference entry (or trim `autosend`/action-respond to a later phase).
3. Then it's mergeable.

The engineering is clean; the gap is that the disclosure didn't move with the capability. Once the metadata tells the truth and Nathan's signed off on the scope, this is good to go. Happy to re-review the moment the permissions block is updated.

santastabber:
Agree with this re-review.

I’m comfortable with Desktop Companion being the first sidecar-class reference entry, but only if the entry is explicit that it is now a **bidirectional control bridge**, not a read-only snapshot bridge.

Before merge, I’d want:

- `extension.json` updated to match the current adapter:
  - reads: `sessions`, `session`
  - writes: `session/draft`, `approval/respond`, `clarify/respond`
  - `webui_navigation: true`
  - owned storage for `hermes-pet-navigation-last-id` and `hermes-pet-action-last-id`
  - an explicit disclosure that the loopback sidecar can issue commands the WebUI adapter executes against the authenticated browser session
- README trust section updated to say the sidecar can switch sessions, inject drafts, optionally autosend, and respond to approval/clarify prompts.
- `scripts/validate-desktop-companion.mjs` updated so this permission drift fails in future.
- `git diff --check` cleaned up; it currently reports trailing blank-line issues in the three pet manifests.

I don’t see this as a reason to reject the architecture outright — the loopback + browser-executed action bridge makes sense for a trusted local desktop companion. But because this PR sets the template for future extension entries, the structured disclosure needs to be accurate before merge.

franksong2702:
Thanks, this is the right call. I pushed `f118fc8` to update the entry disclosure to match the current bidirectional adapter.

Changes made:

- `extension.json` now declares the actual WebUI surface:
  - reads: `sessions`, `session`
  - writes: `session/draft`, `approval/respond`, `clarify/respond`
  - `webui_navigation: true`
  - owned storage: `hermes-pet-navigation-last-id`, `hermes-pet-action-last-id`
  - explicit `sidecar_commands` disclosure for loopback-issued commands that the browser adapter executes in the authenticated WebUI session
- README trust/compatibility sections now state that the local sidecar can switch sessions, inject drafts, optionally autosend, and respond to approval/clarify prompts.
- `scripts/validate-desktop-companion.mjs` now cross-checks the adapter against the permission metadata so this kind of drift fails validation in future.
- Removed the extra trailing blank lines from the three pet manifests.

Validation rerun:

- `node scripts/validate-desktop-companion.mjs`
- `python3 -m json.tool extensions/desktop-companion/extension.json`
- `python3 -m json.tool extensions/desktop-companion/manifest.json`
- `python3 -m json.tool extensions/desktop-companion/pets/*/pet.json`
- `node --check extensions/desktop-companion/assets/companion-adapter.js`
- `git diff --check`

nesquena-hermes:
## Final review of `f118fc8` — disclosure fix verified ✅ (one small read-disclosure gap left)

Re-reviewed the latest push. The blocking finding from my last pass — `extension.json` permissions under-stating the adapter's authority — is **resolved and verified against the code**. The adapter itself is unchanged since `d8bc2bc` (confirmed it's not in the `d8bc2bc...f118fc8` diff), so the security read of the command-execution paths still holds; this push is disclosure + validator only.

**Verified:**
- **Writes now declared and all map to real code:** `session/draft`, `approval/respond`, `clarify/respond` — each `/api/<ep>` confirmed present in `companion-adapter.js`. ✅
- **`webui_navigation: true`**, the two owned storage keys (`hermes-pet-navigation-last-id`, `hermes-pet-action-last-id`), and a new explicit **`sidecar_commands`** block (`can_switch_sessions` / `can_write_drafts` / `can_autosend` / `can_respond_approval` / `can_respond_clarify`) — this is exactly the honest "the loopback sidecar can drive the authenticated session" disclosure the trust model needs. The snapshot's `canReceiveActions` also now reads `true`, consistent. ✅
- **README** spells out the control surface (switch sessions, inject drafts, optionally autosend, respond to approval/clarify). ✅
- **Drift-guard validator** — `node scripts/validate-desktop-companion.mjs` passes on a clean checkout and now cross-checks permissions vs. the adapter, so write-surface drift fails CI in future. Nice — that's a concrete down payment on #6/#8. ✅
- `node --check` clean; all JSON valid; pet-manifest trailing blank lines removed.

### One small residual (non-blocking, but worth tightening on the template entry)

The adapter also issues two **WebUI-origin GET reads** that aren't in the declared `read` set: `/api/approval/pending` and `/api/clarify/pending` (built via the `paths` array around line 673, fetched with `credentials:'include'`). `webui_api.read` currently lists only `["sessions","session"]`, and the README's read bullets don't mention the pending-prompt polls either. Since this is the entry every future submission will copy, please add `approval/pending` + `clarify/pending` to `webui_api.read` (and a line in the README), and — if easy — extend the validator's cross-check to cover declared **reads** too, not just writes (it would have caught this). Strictly a completeness nit; the *write* surface, which is the higher-authority half, is fully disclosed.

### Status

Engineering + disclosure are in good shape. The remaining gate is **not code** — it's the scope call I flagged last round and want to confirm before this lands, since it's the first entry to ship the "extension drives the authenticated session" surface (switch sessions, **autosend messages as the user**, answer approval/clarify gates): **@AvidFuturist / @santastabber — is that bidirectional-control surface in-scope for the first shipped reference entry, or should `autosend` / action-respond be phased behind a later capability?** Once that's confirmed (and ideally the two pending-reads added), I'm comfortable with this merging. Holding the merge for that sign-off rather than calling it myself, since this is the convention-setting entry.

nesquena-hermes:
## My recommendation on the scope call + pre-merge checklist

Following up my review above with a concrete recommendation, since the remaining gate is a scope judgment rather than a code issue. (@AvidFuturist / @santastabber — this is my opinion, not a final ruling; happy to defer if you see it differently.)

### Recommendation: ship the full bidirectional surface — don't defer it — but default-off the two highest-authority behaviors

I'd **approve the bidirectional control surface as-is in capability** (switch sessions, inject drafts, autosend, respond to approval/clarify), rather than phasing `autosend` / action-respond behind future core work. Reasoning:

- **The trust boundary genuinely covers it.** Loopback-only, `credentials:'omit'` to the sidecar, and the sidecar is a localhost process the user chose to run — same trust level as the agent itself, no remote/multi-tenant exposure. Extensions are already "trusted local code running in the WebUI origin against the authenticated session," so writes aren't a new category.
- **It's the actual product, and it's what validates the model.** A desktop pet stripped of autosend/action-respond is a hollow demo, and this is the entry meant to exercise the *full* sidecar-class flow end to end. Curation — a human reading the actual adapter code (done; it's clean: no eval, no external origins, correct credential hygiene) — is the trust mechanism the library is built on.
- **Deferring would over-engineer a single-user, self-hosted surface.** Gating it behind a capability core hasn't shipped would block the flagship entry on infrastructure that isn't the point.

**The one real concern is surprise/consent, not capability:** approval/clarify gates are a human-in-the-loop safety primitive, and autosend speaks *as the user*. Consenting to "run a desktop pet" isn't the same as "let the pet auto-answer my agent's approval prompts." The fix is the *default*, not prohibition:

- **Navigation + draft injection → on by default** (low-risk: you're just changing what you view, or seeing a draft you can still edit/delete before sending).
- **Autosend + approval/clarify-respond → default-off, user opts in** (the "let it act for me" behaviors — off out of the box, enabled deliberately). This gate lives in the **sidecar/native-host** (the WebUI adapter just executes whatever the sidecar sends, so the sidecar must not *issue* those commands until the user enables them).

### ✅ Changes I'd want before merge

1. **(Sidecar / source repo, `franksong2702/hermes-webui-desktop-companion`)** — default `autosend` and approval/clarify-respond **off**; the sidecar only issues those commands once the user explicitly enables them. Navigation + draft injection can stay on by default. Note in the entry README that these two are opt-in.
2. **(Disclosure completeness, this PR)** — add the two WebUI-origin GET reads the adapter performs to `extension.json` → `webui_api.read`: `approval/pending` and `clarify/pending` (currently only `sessions`, `session` are listed), and add a matching line to the README's read bullets.
3. **(Validator, nice-to-have but cheap)** — extend `scripts/validate-desktop-companion.mjs` to cross-check declared **reads** against the adapter too, not just writes — it already guards writes; the missing reads above are exactly what a read-side check would have caught, and it keeps this entry's disclosure self-enforcing as the template for future entries.

Everything else is in good shape and verified: the write-surface disclosure now matches the code, the validator passes, JSON is valid, and the adapter is unchanged from the version I security-reviewed.

Once (1) and (2) are in (3 is optional polish), I'm comfortable with this merging. Pinging @AvidFuturist for the scope sign-off since it's the convention-setting first entry.

santastabber:
> ## My recommendation on the scope call + pre-merge checklist
> 
> Following up my review above with a concrete recommendation, since the remaining gate is a scope judgment rather than a code issue. (@AvidFuturist / @santastabber — this is my opinion, not a final ruling; happy to defer if you see it differently.)
> ### Recommendation: ship the full bidirectional surface — don't defer it — but default-off the two highest-authority behaviors
> 
> I'd **approve the bidirectional control surface as-is in capability** (switch sessions, inject drafts, autosend, respond to approval/clarify), rather than phasing `autosend` / action-respond behind future core work. Reasoning:
> 
>     * **The trust boundary genuinely covers it.** Loopback-only, `credentials:'omit'` to the sidecar, and the sidecar is a localhost process the user chose to run — same trust level as the agent itself, no remote/multi-tenant exposure. Extensions are already "trusted local code running in the WebUI origin against the authenticated session," so writes aren't a new category.
> 
>     * **It's the actual product, and it's what validates the model.** A desktop pet stripped of autosend/action-respond is a hollow demo, and this is the entry meant to exercise the _full_ sidecar-class flow end to end. Curation — a human reading the actual adapter code (done; it's clean: no eval, no external origins, correct credential hygiene) — is the trust mechanism the library is built on.
> 
>     * **Deferring would over-engineer a single-user, self-hosted surface.** Gating it behind a capability core hasn't shipped would block the flagship entry on infrastructure that isn't the point.
> 
> 
> **The one real concern is surprise/consent, not capability:** approval/clarify gates are a human-in-the-loop safety primitive, and autosend speaks _as the user_. Consenting to "run a desktop pet" isn't the same as "let the pet auto-answer my agent's approval prompts." The fix is the _default_, not prohibition:
> 
>     * **Navigation + draft injection → on by default** (low-risk: you're just changing what you view, or seeing a draft you can still edit/delete before sending).
> 
>     * **Autosend + approval/clarify-respond → default-off, user opts in** (the "let it act for me" behaviors — off out of the box, enabled deliberately). This gate lives in the **sidecar/native-host** (the WebUI adapter just executes whatever the sidecar sends, so the sidecar must not _issue_ those commands until the user enables them).
> 
> 
> ### ✅ Changes I'd want before merge
> 
>     1. **(Sidecar / source repo, `franksong2702/hermes-webui-desktop-companion`)** — default `autosend` and approval/clarify-respond **off**; the sidecar only issues those commands once the user explicitly enables them. Navigation + draft injection can stay on by default. Note in the entry README that these two are opt-in.
> 
>     2. **(Disclosure completeness, this PR)** — add the two WebUI-origin GET reads the adapter performs to `extension.json` → `webui_api.read`: `approval/pending` and `clarify/pending` (currently only `sessions`, `session` are listed), and add a matching line to the README's read bullets.
> 
>     3. **(Validator, nice-to-have but cheap)** — extend `scripts/validate-desktop-companion.mjs` to cross-check declared **reads** against the adapter too, not just writes — it already guards writes; the missing reads above are exactly what a read-side check would have caught, and it keeps this entry's disclosure self-enforcing as the template for future entries.
> 
> 
> Everything else is in good shape and verified: the write-surface disclosure now matches the code, the validator passes, JSON is valid, and the adapter is unchanged from the version I security-reviewed.
> 
> Once (1) and (2) are in (3 is optional polish), I'm comfortable with this merging. Pinging @AvidFuturist for the scope sign-off since it's the convention-setting first entry.



This sounds right to  #me.

I’m good with Desktop Companion being the first sidecar-class reference entry, including the bidirectional bridge. I don’t think we need to defer that behind future proxy work.

My only requirement is that the “acts as the user” parts are not surprising: autosend and approval/clarify responses should be opt-in/default-off in the sidecar/native host. Navigation and draft injection are fine by default.

So I’d just like the pending read disclosures added here (`approval/pending`, `clarify/pending`) and the opt-in/default-off behavior documented clearly. After that I’m comfortable merging. The validator read-check would be a nice follow-up if it’s easy.

franksong2702:
Updated this PR for the review feedback:

- Declared the adapter pending-context reads: `approval/pending` and `clarify/pending`.
- Documented that navigation + draft injection are available by default, while direct autosend and inline approval/clarify responses are default-off and require local user opt-in.
- Extended the Desktop Companion validator to check the new read disclosures and the adapter pending endpoints.

I also opened the matching source-side PR for the actual sidecar/native-host gates and pet UI controls: https://github.com/franksong2702/hermes-webui-desktop-companion/pull/4

santastabber:
Thanks Frank — this is the shape I was looking for.

The read disclosures and validator coverage here look like the right fix for the extension-entry side. Since the default-off behavior is actually enforced in the companion source repo, I’d like to see the source-side PR land or be verified before merging this entry.

Once the sidecar/native-host gates are confirmed, I’m comfortable with #11. Screenshots / packaging-size questions can stay follow-ups.

franksong2702:
Source-side follow-up is now updated and verified in franksong2702/hermes-webui-desktop-companion#4.

That PR now includes the actual sidecar/native-host gates for the default-off behavior, plus the Desktop Pet confirmation-card UI and right-click Permission control menu. I also pushed the confirmation-card polish after local testing on the 8787 extension flow.

Verified locally:

- `npm test` in the source repo → 28/28 passed
- `cargo check` in `desktop-pet/src-tauri` → passed
- 8787 Desktop Pet permission confirmation card manual check → passed

Extension-entry side remains at this PR head with the pending read disclosures and validator coverage already added.

franksong2702:
Source-side PR #4 has now landed in `franksong2702/hermes-webui-desktop-companion`.

That merge includes the default-off sidecar/native-host gates, the Desktop Pet first-use confirmation cards, the right-click Permission control menu, and the confirmation-card UI polish we verified locally on the 8787 extension flow.

So the source-side requirement Santa called out is now satisfied from my side.

nesquena-hermes:
## Verified — all pre-merge items addressed ✅

Re-checked `5ec7772` on a clean checkout (PR head unchanged since my last pass, so the security review of the adapter still holds). Everything from the checklist is in, @santastabber signed off on the scope, and the source-side gate has now merged too:

- **Pending reads disclosed** — `approval/pending` + `clarify/pending` added to `webui_api.read` and the README read bullets. I re-reconciled the adapter's full WebUI-origin surface against the declared permissions and there are now **zero gaps**: reads = `sessions`, `session`, `approval/pending`, `clarify/pending`; writes = `session/draft`, `approval/respond`, `clarify/respond`; the `/api/pet/*` + `/api/webui/snapshot` calls are loopback (sidecar origin), correctly disclosed via `loopback_sidecar` + `sidecar_commands` rather than as WebUI reads.
- **Opt-in / default-off documented** — the README states navigation + draft injection are default-on, while autosend and inline approval/clarify responses are **default-off and require local user opt-in**; the metadata also encodes `autosend_default: false` + `approval_clarify_response_default: false`, so the "acts as the user" behaviors are explicit in the entry, not just prose.
- **Validator** — now cross-checks the declared reads (and that the adapter actually references the pending endpoints), not just writes. Passes clean (`node scripts/validate-desktop-companion.mjs` → exit 0), JSON valid, `node --check` clean.
- **Source-side enforcement landed** — the default-off gates, first-use confirmation cards, and the right-click permission control menu merged in `franksong2702/hermes-webui-desktop-companion#4` (`9553bf5a`). That was the one thing the adapter itself can't enforce (it only executes what the sidecar issues), so it correctly lives in the sidecar/native host — and it's now shipped, not pending.

### Overall

This is a clean, well-disclosed sidecar-class reference entry, and it's the first to exercise the full "extension drives the authenticated session" flow end to end. The trust posture is honest and the high-authority behaviors are off-by-default with explicit opt-in. **From my side there's nothing left to change** — code, disclosure, validator, and the source-side gate are all in good shape, and the scope is signed off.

What's remaining before merge is now purely a maintainer's go: per our process I'm not self-merging the convention-setting first entry, so this is ready for @AvidFuturist / @santastabber to merge whenever you're ready. Great work, @franksong2702 — this sets a strong template for the library.

Issue #12, opened 2026-06-23
🗺️ Extension system roadmap (umbrella / tracking)


nesquena-hermes:
# 🗺️ Extension system roadmap — umbrella / tracking issue

The single place to see the whole extension-system effort: what's done, what's left, which repo each piece lands in, and the build order. Full design rationale lives in [`VISION.md`](../blob/main/VISION.md) and [RFC 0001](../blob/main/docs/rfcs/0001-extension-json-schema.md).

## The goal

Core WebUI stays lean and curated; everything good-but-out-of-scope-for-core becomes a **vetted extension** in this library, browsable and one-click-installable from inside WebUI (Settings → Extensions). "In the registry == vetted." Extensions are **trusted local code** (extension JS runs in the WebUI origin against the authenticated session), so curation + automated safety gates are the load-bearing trust boundary.

## Two-repo split (which repo implements what)

- **`nesquena/hermes-webui` (core)** — the extension *infrastructure*: loader, manifest contract, same-origin asset serving, the Settings → Extensions UI, the install client, and the consented sidecar proxy.
- **`hermes-webui/hermes-webui-extensions` (this repo, library)** — the *entries*, conventions, schema, the registry generator, entry-PR CI + safety gates.

## ✅ Foundation phase — DONE (merged 2026-06-23)

- [x] Extension loader + manifest bundles + sandboxed `/extensions/*` serving + `/api/extensions/status` diagnostics — **core** (shipped v0.51.545–.553).
- [x] Settings → Extensions: diagnostics pane + loopback-sidecar health card + **enable/disable installed extensions** — **core** (v0.51.553 / #4612 / #4637).
- [x] Library foundation — README / CONTRIBUTING / entry contract / examples (#1).
- [x] **VISION.md** roadmap (#2, merged).
- [x] **RFC 0001 — `extension.json` schema** (#10, merged): one-file model, core-owned capabilities, purpose-based permissions, split lifecycle.
- [x] **First real entry: Desktop Companion** (#7, closed → PR #11 merged) — first sidecar-class reference entry; bidirectional control with the "acts as the user" behaviors default-off/opt-in, enforced in the sidecar. Ships a per-entry validator that cross-checks declared permissions vs. the adapter.

## ✅ Registry → gallery → one-click-install chain — SHIPPED (2026-06-25)

[...]

PR #14, opened 2026-06-24
Add extension delivery artifacts

franksong2702:
## Summary

- generate deterministic per-extension zip artifacts under `dist/artifacts/`
- add `download`, artifact-level `sha256`, and `artifact_size` to generated registry entries
- publish the whole `dist` directory through the existing GitHub Pages workflow
- document the boundary: this repo produces reviewed artifacts; core WebUI still owns fetch, verify, safe extract, rollback, and uninstall

## Scope

This is the library-side delivery piece for #9. It does not implement the core Settings -> Extensions gallery/install client; that remains tracked in nesquena/hermes-webui#4746. It also does not change the sidecar proxy work in #5 / nesquena/hermes-webui#4747.

Closes #9.
Refs #12.

## Verification

- `node --check scripts/extension-registry-lib.mjs scripts/generate-registry.mjs scripts/test-extension-validator.mjs scripts/validate-extensions.mjs scripts/validate-desktop-companion.mjs`
- `node scripts/validate-extensions.mjs`
- `node scripts/test-extension-validator.mjs`
- `rm -rf dist && node scripts/generate-registry.mjs --out dist/registry.json`
- `unzip -l dist/artifacts/desktop-companion-0.1.0.zip`
- verified generated registry hash matches the zip bytes

PR #15, opened 2026-06-24
Add extension safety gates

franksong2702:
## Summary

Adds the first automated safety-gate layer for curated extension entries.

- add `scripts/scan-extension-safety.mjs` to fail closed on high-risk entry shapes: obvious secrets, symlinks or unsafe paths, dangerous JavaScript execution patterns, undeclared external URL literals, undeclared localStorage writes, and generated artifact integrity drift
- run the safety scan in the Extensions GitHub Action before registry generation
- cover cwd-safe execution from `scripts/` in the validator self-test
- document the new local validation command and safety-gate scope

## Validation

- `node --check scripts/extension-registry-lib.mjs && node --check scripts/generate-registry.mjs && node --check scripts/test-extension-validator.mjs && node --check scripts/validate-extensions.mjs && node --check scripts/validate-desktop-companion.mjs && node --check scripts/scan-extension-safety.mjs`
- `rm -rf dist dist-from-scripts && node scripts/generate-registry.mjs --out dist/registry.json`
- from `scripts/`: `node validate-extensions.mjs && node scan-extension-safety.mjs && node generate-registry.mjs --out ../dist-from-scripts/registry.json`
- `git diff --check`

Addresses the first automated malicious-code / artifact-integrity gate for #8. Cross-reference: #12.

### DIFF
--- a/.github/workflows/extensions.yml
+++ b/.github/workflows/extensions.yml
@@ -42,6 +42,9 @@ jobs:
       - name: Test extension validator
         run: node scripts/test-extension-validator.mjs
 
+      - name: Scan extension safety gates
+        run: node scripts/scan-extension-safety.mjs
+
       - name: Generate registry
         run: node scripts/generate-registry.mjs --out dist/registry.json
 
--- a/README.md
+++ b/README.md
@@ -92,15 +92,16 @@ Run the current repo-wide checks locally with:
 ```bash
 node scripts/validate-extensions.mjs
 node scripts/test-extension-validator.mjs
+node scripts/scan-extension-safety.mjs
 node scripts/generate-registry.mjs --out dist/registry.json
 ```
 
-Pull requests run the same validator in CI. Pushes to `main` generate the
-registry and per-extension zip artifacts for GitHub Pages. The registry entry
-includes a `download` URL and artifact-level `sha256` for each extension so the
-future WebUI install client can fetch and verify reviewed bytes before
-extracting them. The core-side install client, safe extraction, rollback, and
-uninstall flow are tracked in the main Hermes WebUI repo.
+Pull requests run the same validator and safety scan in CI. Pushes to `main`
+generate the registry and per-extension zip artifacts for GitHub Pages. The
+registry entry includes a `download` URL and artifact-level `sha256` for each
+extension so the future WebUI install client can fetch and verify reviewed
+bytes before extracting them. The core-side install client, safe extraction,
+rollback, and uninstall flow are tracked in the main Hermes WebUI repo.
 
 ## Current Entries
 
--- a/docs/extension-entry.md
+++ b/docs/extension-entry.md
@@ -130,12 +130,18 @@ Run the repo-wide validator before opening or updating an extension PR:
 
 ```bash
 node scripts/validate-extensions.mjs
+node scripts/scan-extension-safety.mjs
 ```
 
 The validator scans every `extensions/*/extension.json`, checks required files,
 safe local asset paths, runtime `manifest.json` consistency, shipped capability
 names, lifecycle and permission shape, and selected permissions-vs-code drift
 such as WebUI API read/write disclosures.
+
+The safety scan layers on high-risk checks for entry files before they can land:
+obvious secrets, symlinks or unsafe paths, blocked JavaScript execution patterns,
+undeclared external network literals, localStorage writes without owned-key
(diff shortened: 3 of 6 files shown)


nesquena-hermes:
## Review — approving ✅ (empirically verified on a clean checkout of head `dc46db9`)

Thanks @franksong2702 — this is a clean, well-scoped first automated safety-gate layer for #8, and it even repairs a latent false-green from the earlier validator work. I reviewed it as convention-defining infrastructure (the tooling every future entry gets held to), so I ran every gate rather than reading the prose.

### What I verified (all green)

**Syntax + real-repo runs**
- `node --check` on all 6 `.mjs` scripts — pass
- `validate-extensions`, `test-extension-validator` (incl. the new cwd-safe scan self-test case), and `scan-extension-safety` — all green against the real `extensions/*`

**The gates actually fire** (injected fixtures on an otherwise-valid entry, then reverted):
- committed secret (AWS-key shape) → caught
- `document.cookie` → caught — and notably the **base validator passes this (rc=0)** while the safety scan blocks it, so this is real added coverage, not just overlap
- `process.env`, `node:child_process`, symlinked entry file → all caught
- artifact integrity drift (sha256 / size / zip-magic mismatch) → covered by `scanArtifacts()`

**cwd-safety** — runs correctly from repo root, from `scripts/`, and from an unrelated cwd (`/tmp`). It resolves the scan root from the script location, not `process.cwd()`, which **fixes the subdir false-green** the earlier validator class had. Nice.

**Determinism + integrity** (delivery layer it re-verifies):
- registry generated twice → zip **byte-for-byte identical** (`cmp -s`), registry identical modulo `generated_at`/`published_at`
- `sha256`, `artifact_size`, `file_count == len(file_sha256)` all match the on-disk bytes
- zip is a real archive (`unzip -l` reads it), fixed 1980 timestamps, every member rooted under `desktop-companion/` (zip-slip-safe)

**CI wiring** — the scan sits at the right point in the sequence: `validate → test-validator → safety-scan → generate → validate-JSON → upload/deploy (main-gated)`, so it **fails closed before anything is generated or published**. Workflow permissions stay least-priv (`contents:read` top-level; `pages:write`+`id-token:write` only on the deploy job). PR check is green; deploy correctly skipped on the PR. No new deps (pure `node:` stdlib). No RFC-0001 drift — the doc edits accurately describe what ships.

### Non-blocking notes (scope-appropriate, no changes required)

1. **Regex gates have expected bypasses** — I confirmed a split-string `XMLHttpRequest` URL (`'https'+'://…'`) and an aliased `Function` constructor (`globalThis['Fun'+'ction']`) slip through. That's *correct* for a lightweight fail-closed CI gate; deeper malicious-intent analysis is the rest of #8's future work. Worth a one-line code comment so a green scan is never mistaken for a security guarantee.
2. **Secret scan reads an entry's own docs/README**, so an entry that documents an example token could false-positive. Rare, and the strictness is defensible — just a latent FP to be aware of.
3. **No Zip64 guard** (inherited from the #14 zip writer, not introduced here): 32-bit size / 16-bit count fields mean a >4 GB file or >65535-file entry would overflow. Realistic entries are MB-scale so it's fine today; a loud throw-on-oversize would be good hygiene since this is the format every entry inherits.

### Verdict

Code-ready, no fixes needed. Merging as the first safety-gate layer for #8. cc @santastabber

🤖 Reviewed empirically (clean checkout + injected-fixture gate tests + determinism/integrity verification) per the extensions-repo infrastructure-PR review process.

nesquena-hermes:
Correction to nit #3 above (no Zip64 guard) — that's **stale**, my apologies. I was reading the diff in isolation; the guard already landed in #14 (the zip producer this scan re-verifies). Merged `main` has it: `ZIP32_MAX`/`ZIP16_MAX` constants with loud `throw`s on oversized members, central-directory offset/size overflow, and >65535-file counts — so the writer fails closed rather than emitting a corrupt archive. So that nit is resolved; only nits #1 (regex bypasses are by-design for a lightweight gate — worth an in-code comment) and #2 (docs-with-example-tokens could false-positive) stand, both non-blocking.

Merged as commit `0003990` — the post-merge CI run logged `safety scan passed for 1 extension entry`, then regenerated + redeployed the registry; the live published zip's sha256 still matches the registry trust anchor. Issue #8 + the umbrella #12 updated to reflect the first fail-closed safety-gate layer. Thanks again @franksong2702.

PR #19, opened 2026-06-28
Add message-pins extension (client-side message pinning)

nesquena-hermes:
## Summary

Adds a **`message-pins`** extension: pin individual messages in a conversation.

- Per-message **pin button** in each message's hover action bar (with a floating fallback for rows that have no action bar).
- A **header button** near the top-right of the chat area with a **count badge**; clicking it opens a **popover** listing the pinned messages, each with **click-to-jump** and an unpin control.
- **3-pin cap** per conversation with a polite over-cap toast.
- **Client-side persistence** via `localStorage`, keyed per `session_id`, so each conversation keeps its own pins and switching conversations swaps them.
- A **`MutationObserver`** re-applies pin decorations after the transcript re-renders (the message list rebuilds `innerHTML`), so pins survive streaming, scrolling, and reloads.

This is a `static-ui` / `manifest-bundle` entry: no backend routes, no sidecar, no network, no filesystem, no core changes. The only state it writes is the single `localStorage` key `hermes-ext-message-pins`.

## Design reference / credit

The pin UX here follows closed core PR **nesquena/hermes-webui#2534** by **@Michaelyklam** (originally requested in issue #2508) — the per-message pin button, header popover with badge, click-to-jump, and 3-pin cap are his design. That PR implemented pinning with a new core API and server-side persistence, which didn't fit core's scope; this reimagines it as an opt-in extension that persists pins in the browser and needs no core changes. Co-authored to preserve attribution. Thanks @Michaelyklam.

## Verification

Repo gates (all green, 3 entries now):

```
node scripts/validate-extensions.mjs       # validated 3 extension entries
node scripts/scan-extension-safety.mjs     # safety scan passed for 3 extension entries
node scripts/test-extension-validator.mjs  # self-tests passed
node scripts/generate-registry.mjs --out dist/registry.json   # wrote 3 entries + artifacts
node --check extensions/message-pins/assets/message-pins.js
```

**Live end-to-end test** (isolated WebUI server with the extension loaded via `HERMES_WEBUI_EXTENSION_DIR` + `HERMES_WEBUI_EXTENSION_MANIFEST`, seeded session with 8 messages, driven via headless Chrome/CDP):

- extension boots; header button injected; **8 message rows → 8 pin buttons**; session id read correctly from the DOM
- pinning messages 0/2/4 → stored `[0,2,4]`, badge `3`, 3 active buttons
- 4th pin → **capped**, stays `[0,2,4]`, polite toast shown
- popover opens, lists 3 pins with correct message-text previews
- click-to-jump scrolls to the row and flashes it, closing the popover
- **reload** → pins persist (`[0,2,4]`), buttons re-decorate, badge `3` (MutationObserver re-apply confirmed)
- unpin via popover → `[2,4]`, badge `2`
- forced transcript re-render → all pin buttons re-applied, active state preserved
- assets served 200, zero extension errors in the server log

- `webui_api.read/write: []` — no `fetch`, no `/api/*` calls
- `network_external: false` — no external URLs
- `storage.owned: ["hermes-ext-message-pins"]` — the one and only localStorage key written
- `dom.owned: true`, `dom.mutates_core_views: true` — injects the pin buttons / header button / popover and toggles its own decoration classes on message rows
- `webui_navigation: false` — scrolls within the transcript only; no panel/route navigation
- no eval, no cookies, no loopback/native host; all message text rendered via `textContent` (XSS-safe)

## Notes

- Pins are per-browser (client-side only), not synced across devices — the deliberate trade-off for a no-backend extension.
- Jump-to-pin requires the target message to be within the loaded transcript window; for a very long conversation an off-window pin shows a "scroll up to load it" notice rather than silently doing nothing.
- Relies on the current message-row DOM (`data-msg-idx`, `.msg-actions`, `data-session-id`) until a formal message-decoration extension API exists.

### DIFF
--- a/extensions/README.md
+++ b/extensions/README.md
@@ -21,3 +21,5 @@ need it and maintainers agree on the shared contract.
   sidecar-class extension candidate.
 - `mobile-conversations`: phone-only floating Conversations button, same-location
   drawer close X, and long-press shortcuts for the existing Hermes WebUI mobile drawer.
+- `message-pins`: pin individual messages in a conversation, with a header
+  popover, click-to-jump, and client-side per-session persistence.
--- a/extensions/message-pins/README.md
+++ b/extensions/message-pins/README.md
@@ -1,0 +1,181 @@
+# Message Pins
+
+Message Pins is a trusted local Hermes WebUI extension that lets you pin
+individual messages in a conversation. It adds a pin button to each message, a
+header popover that lists your pinned messages with click-to-jump, and a small
+badge showing the pin count. Pins are saved locally in your browser, per
+session.
+
+## What It Does
+
+- Adds a pin button to each message (in the hover action bar, with a floating
+  fallback for messages that have no action bar).
+- Pinning a message records it; pinning is capped at 3 messages per conversation
+  with a polite over-cap notice (unpin one to pin another).
+- Adds a pin button near the top-right of the chat area with a count badge.
+  Clicking it opens a popover listing the pinned messages.
+- Clicking a pinned message in the popover scrolls to it and briefly highlights
+  it. If that message has been virtualized out of the loaded transcript window,
+  the extension tells you to scroll up to load it first rather than silently
+  doing nothing.
+- Each popover entry has an unpin control.
+- Pins are stored per `session_id`, so each conversation keeps its own set, and
+  switching conversations swaps the displayed pins.
+
+## Current Shape
+
+```text
+Hermes WebUI page
+  -> manifest-bundled extension assets
+  -> /extensions/assets/message-pins.js
+  -> /extensions/assets/message-pins.css
+  -> existing WebUI message rows ([data-msg-idx]) + .messages-shell
+  -> localStorage (key: hermes-ext-message-pins)
+```
+
+This extension is `static-ui` / manifest-bundle only. It does not add backend
+routes, start a sidecar, access external networks, read or write files, or use
+native host APIs. Persistence is client-side `localStorage`, so pins are
+per-browser and not synced across devices.
+
+## Capabilities
+
+This entry declares only capabilities currently available to extension entries:
+
+- `manifest-bundle`
+
+It does not require `loopback-sidecar` or future sidecar proxy support.
+
+## Install For Local Testing
+
+Start Hermes WebUI with this extension directory:
+
+```bash
+cd /path/to/hermes-webui
+HERMES_WEBUI_EXTENSION_DIR=/path/to/hermes-webui-extensions/extensions/message-pins HERMES_WEBUI_EXTENSION_MANIFEST=manifest.json ./start.sh
+```
+
+The runtime `manifest.json` injects:
+
+```json
+{
+  "extensions": [
+    {
+      "id": "message-pins",
+      "name": "Message Pins",
+      "description": "Pin individual messages in a conversation, with a header popover and click-to-jump.",
+      "scripts": ["assets/message-pins.js"],
+      "stylesheets": ["assets/message-pins.css"]
+    }
+  ]
+}
+```
+
+Once the in-app extensions gallery is available, this entry can also be browsed
+and installed from Settings -> Extensions.
+
+## Disable And Uninstall
+
+To disable the WebUI extension, restart Hermes WebUI without:
+
+```text
+HERMES_WEBUI_EXTENSION_DIR
+HERMES_WEBUI_EXTENSION_MANIFEST
+HERMES_WEBUI_EXTENSION_STYLESHEET_URLS
+HERMES_WEBUI_EXTENSION_SCRIPT_URLS
+```
+
+To uninstall it from a manual local setup, remove the
+`extensions/message-pins/` directory from the local extensions checkout or
+remove the entry from the aggregate manifest that points at it.
+
+Your saved pins live under the `hermes-ext-message-pins` key in your browser's
+localStorage. Clearing that key (or your site data) removes all saved pins.
+
+## Trust And Permissions
+
+This is trusted local code. The injected JavaScript runs in the Hermes WebUI
+browser origin and can use the logged-in browser session.
+
+Current disclosed behavior:
+
+- creates extension-owned DOM for the pin buttons, header button, popover, and a
+  small toast
+- inserts the pin button into each message's existing `.msg-actions` bar, with a
+  floating fallback appended to the message row when no action bar exists
+- inserts the header button into the existing `.messages-shell`
+- reads message rows via the existing `data-msg-idx`, `data-raw-text`, and
+  `data-session-id` attributes already present on the page
+- reads the active session id from a rendered message's `data-session-id` (or
+  from the `/session/<id>` URL as a fallback)
+- uses a `MutationObserver` on the messages container to re-apply pin
+  decorations after the transcript re-renders
+- reads and writes `localStorage` under the single key `hermes-ext-message-pins`
+- does not call WebUI HTTP APIs
+- does not access cookies
+- does not contact loopback or external network services
+- does not use arbitrary filesystem or native host APIs
+
+All message text rendered into the popover and previews is inserted via
+`textContent` (or escaped), so message content cannot inject markup.
+
+## Compatibility
+
+Required WebUI surface:
+
+- manifest-bundled extension assets
+- same-origin extension asset serving under `/extensions/`
+- message rows carrying `data-msg-idx` (and, where available, `data-raw-text`
+  and `data-session-id`), inside `#messages` / `.messages-shell`
+
+This entry targets Hermes WebUI builds that render message rows with
+`data-msg-idx` and the `.msg-actions` hover bar. The compatibility requirement is
+better described as the manifest bundle plus the message-row DOM surface above.
+
+## Verification
+
+From this repository:
+
+```bash
+node scripts/validate-extensions.mjs
+node scripts/scan-extension-safety.mjs
+node scripts/generate-registry.mjs --out dist/registry.json
+node --check extensions/message-pins/assets/message-pins.js
+python3 -m json.tool extensions/message-pins/extension.json
+python3 -m json.tool extensions/message-pins/manifest.json
+```
+
+Manual verification should confirm:
+
+- hovering a message shows a pin button; clicking it pins the message and the
+  header badge count increases
+- the header pin button opens a popover listing the pinned messages
+- clicking a pinned entry scrolls to and highlights that message
+- the unpin control in the popover removes a pin
+- pinning a 4th message shows the over-cap notice instead of pinning
+- reloading the page keeps the pins for that conversation
+- switching to another conversation shows that conversation's own pins
+- a pin whose message is scrolled out of the virtualized window shows the
+  "scroll up to load it" notice instead of doing nothing
+
+## Known Limitations
+
+- Pins are stored per browser via `localStorage`; they are not synced across
+  devices or browsers (a deliberate consequence of being a client-side-only
+  extension with no backend).
+- The extension relies on the current message-row DOM (`data-msg-idx`,
+  `.msg-actions`) until a formal message-decoration extension API exists.
+- Jump-to-pin requires the target message to be within the loaded transcript
+  window; very old messages in a long conversation may need to be scrolled into
+  range first.
+
+## Credit
+
+The pin UX in this extension — a per-message pin button, a header popover with a
+count badge, click-to-jump, and a 3-pin cap — follows the design from closed
+core PR [#2534](https://github.com/nesquena/hermes-webui/pull/2534) by
+**@Michaelyklam** (originally requested in issue #2508). That PR implemented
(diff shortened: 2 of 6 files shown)


franksong2702:
I ran the repo gates locally:

```bash
node scripts/validate-extensions.mjs
node scripts/scan-extension-safety.mjs
node scripts/test-extension-validator.mjs
node scripts/generate-registry.mjs --out dist/registry.json
node --check extensions/message-pins/assets/message-pins.js
python3 -m json.tool extensions/message-pins/extension.json
python3 -m json.tool extensions/message-pins/manifest.json
```

All passed. The permission disclosure also matches the code on the security-sensitive parts I checked: no `fetch`, no `/api/*`, no cookies, no external/loopback URLs, and the only storage write is `localStorage.setItem('hermes-ext-message-pins', ...)`. Message previews render through `textContent`, and the remaining `innerHTML` writes are static SVG/icon markup.

One issue I would fix before merge:

`extensions/message-pins/assets/message-pins.css:54-68` positions the new header control at `top: 12px; right: 16px; width: 32px; height: 32px; z-index: 11`, appended to `.messages-shell` from `message-pins.js:242-262`.

Core WebUI already has `#jumpToSessionStartBtn` in the same `.messages-shell` (`static/index.html:425`) and styles it as a top-right control: `.session-jump-btn { right:20px; height:32px; z-index:11; }` plus `.session-jump-btn--start { top:16px; }` (`static/style.css:1723-1725`). In long conversations where session navigation is enabled and the Start button is visible, these occupy the same corner and one control can hide or intercept the other.

Recommended fix: reserve a separate slot for the pins header button. For example, place it to the left of `#jumpToSessionStartBtn` when that button exists/is visible, or compute the `right` offset from that button's bounding box. A simpler first pass is also fine if it is deterministic: choose a non-overlapping top-right offset and test with `messages.session-nav-enabled` plus the Start button visible.

I do not see a backend/security blocker in the extension shape. The DOM positioning collision is the merge-risk I would address.

franksong2702:
Follow-up from real 8787 testing in a long Hermes WebUI conversation:

The basic persistence path appears to work when clicking a truly visible inline pin button: in my local browser automation, the pin was written to `localStorage`, the per-message button became active, and the header badge updated from 0 to 1.

However, the current UX still feels broken in real manual testing:

1. The header pin button overlaps the core session navigation control. The extension positions `.hwx-pin-header-btn` at `top: 12px; right: 16px; z-index: 11`, while core WebUI's `#jumpToSessionStartBtn` / `.session-jump-btn--start` is also positioned at the top right (`top: 16px; right: 20px; z-index: 11`). On long conversations, the pin button visually collides with the built-in "Start" button.

2. The row decoration target is too broad. The extension scans all `#messages [data-msg-idx]`, which includes empty assistant anchor/worklog segment nodes such as `.assistant-segment-worklog-source.assistant-segment-anchor`. In a long conversation this creates many hidden/floating pin buttons on non-message anchor nodes before the actual visible message/action rows. That makes the DOM noisy and can make the interaction feel like it is not doing anything.

3. Manual tester feedback was: the per-message Pin button is visible, but clicking it appears to do nothing, and the top-right Pin icon is duplicated/overlapping with another control. I think this is mainly because the successful state is hard to discover once the header entry is overlapped, and the active feedback on the message button is subtle.

I would treat this as not ready to merge yet. Suggested fix direction:

- reserve a non-overlapping slot for the header/list affordance, or integrate it with the existing session-nav positioning instead of hard-coding the same top-right corner
- decorate only real, visible message rows/assistant content segments with usable text/action bars, not empty anchor segments
- make the successful pinned state more obvious after click, especially when the pinned-list affordance is not already open

nesquena-hermes:
Thanks for the thorough real-conversation testing, @franksong2702 — this was exactly the right way to catch these, and all three issues are now fixed on the branch. Recap of what changed and how I verified each against your feedback:

## 1. Header button collided with the core session-nav "Start" button

You were right that `.hwx-pin-header-btn` hard-coded `top:12px; right:16px` into `.messages-shell`, the same top-right corner as core's `#jumpToSessionStartBtn` / `.session-jump-btn--start` (`top:16px; right:20px; z-index:11`), which only appears in long, session-nav-enabled conversations.

**Fix:** the position is now computed in JS (`positionHeaderButton()`), exactly as you suggested. When the Start button is visible, the pins button is placed to its **left**, offset computed from the Start button's bounding box (`right = 20 + startBtn.width + 8px gap`); otherwise it uses the default top-right slot. It recomputes on the Start button's attribute flips (a `MutationObserver` on its `style`/`class`) **and** on `.messages-shell` scroll, since the Start button toggles on scroll without a `#messages` mutation. The hard-coded `top/right` were removed from the CSS so the JS values win.

**Verified** in a live 16-message, session-nav-enabled conversation with the Start button visible: the two rects no longer intersect (`overlap: false`) — pin button at right≈90px, sitting cleanly to the left of Start with an 8px gap.

## 2. Decoration target was too broad (anchor/worklog segments)

Confirmed: `redecorate()` scanned all `#messages [data-msg-idx]`, which includes the hidden `assistant-segment-anchor` / `assistant-segment-worklog-source` nodes (both `display:none`), producing floating pin buttons on non-message nodes and making the interaction feel dead.

**Fix:** added `isRealMessageRow()` — it skips those anchor/worklog segment classes, requires a real content surface (`.msg-body` or `.msg-foot .msg-actions` / `.msg-actions`), and requires `getBoundingClientRect().height > 1` (excludes the `display:none` anchors and collapsed nodes). Only real, visible message rows are decorated.

**Verified:** injected 3 synthetic noise nodes (an `assistant-segment-anchor`, an `assistant-segment-worklog-source`, and a zero-height empty `[data-msg-idx]`) into the live transcript and re-ran decoration — raw `[data-msg-idx]` count went 16 → 19, but pin buttons stayed at **16** and **0 noise nodes were decorated**.

## 3. Successful pinned state was too subtle / "clicking appears to do nothing"

This was largely a compound of #1 and #2 (the success cue was hidden behind the overlapped header, and the noise buttons made clicks feel inert), plus the active state being just a color swap.

**Fix:** the active pin button is now a clearly **filled accent chip** (background + contrast color), not a color swap, and a pinned row gets a visible **accent left-rail marker** so a pinned message is obvious at a glance even when the pins popover isn't open. The "Message pinned" toast remains.

**Verified:** clicking a real inline pin button produces an active filled button (non-transparent background confirmed via computed style), a `.hwx-pinned-row` marker, and the header badge updating to `1` — captured in a screenshot of a long conversation showing the pin button + count badge sitting beside (not overlapping) the Start button, with the toast visible.

## Unchanged / still green

Your gate run and permission-disclosure review still hold: gates pass (`validate` / `safety-scan` / `validator self-test`), and the security shape is unchanged — no `fetch`, no `/api/*`, no cookies, no external/loopback URLs, the only storage write is `localStorage['hermes-ext-message-pins']`, previews via `textContent`, and `innerHTML` only for static SVG.

More broadly: your review changed how I'm testing every extension — I now test each DOM-injecting extension in a long, realistic, session-nav-enabled conversation (with synthetic anchor/worklog noise injected) and vision-check for overlap/duplication/floating controls, rather than verifying mechanism in a short seeded session. That pass also caught and fixed an unrelated rail-width assumption in the external-app-tab extension. Thanks again — this was a genuinely valuable review.

nesquena-hermes:
@rodboev — tagging you for a second look here too. @franksong2702 already gave this a thorough real-conversation review (thank you — see the recap above), and all three issues he found are now fixed and verified:

- **header button no longer collides** with core's `#jumpToSessionStartBtn` (position is computed from that button's bounding box and recomputes on scroll / its attribute flips)
- **decoration now skips hidden anchor/worklog segments** (real-row guard: content surface + non-zero height) — verified 0 noise nodes decorated after injecting synthetic ones
- **success state is now obvious** (filled accent chip + pinned-row accent rail)

Re-verified in a live 16-message, session-nav-enabled conversation with a vision check (pin button sits cleanly left of the Start button, no floating buttons, toast + badge visible).

Both of you: a hands-on browser pass in a long conversation is the gate that matters most here — if anything still feels off in real use, please flag it. Gates green, rebased on main.

franksong2702:
Follow-up from another real 8787 pass on latest PR head (`fe0d9b0...`), using a long real Hermes WebUI conversation.

The previous fixes addressed the header overlap and bogus hidden-row decoration. I hit two more interaction issues while manually testing the pinned-message flow:

## 1. Inline pin/unpin still feels unreliable to click

Observed behavior: the first pin can succeed, but clicking the same inline pin again to unpin can feel delayed or dead. A second attempt may appear to do nothing.

Root cause from hit-testing in the real page:

- For assistant turns whose footer contains the core question-jump affordance, core CSS intentionally keeps `.msg-actions` quiet until hover/focus and sets that action area to `pointer-events: none` at rest.
- The extension inserts `.hwx-pin-btn` inside that `.msg-actions` container. Without an explicit override, the visible pin icon can be shown but the click target can resolve to `.msg-foot` instead of the button.
- After pinning, the active inline button also changes dimensions (`23x17` to about `21x17` in my run). Combined with the tiny hit target and SVG child being the hit target, manual unpin feels easy to miss.

Local patch that fixed it for testing:

```css
.hwx-pin-btn {
  position: relative;
  z-index: 2;
  pointer-events: auto;
  flex: 0 0 auto;
  width: 24px;
  min-width: 24px;
  height: 24px;
  padding: 0;
  line-height: 1;
  border-radius: 6px;
}
.hwx-pin-btn svg { pointer-events: none; }
.hwx-pin-btn--inline.hwx-pin-btn--active { padding: 0; }
```

Verification after the local patch: with real `page.mouse` clicks on the visible inline button, I could toggle the same message 5 times in a row (`pin / unpin / pin / unpin / pin`). The button rect stayed stable at `24x24`, `elementFromPoint()` hit the `BUTTON`, and `localStorage['hermes-ext-message-pins']` alternated correctly.

## 2. Popover click-to-jump centers long messages

Observed behavior: clicking a pinned entry in the top-right popover scrolls to an odd position for long assistant/final-answer messages. It lands around the middle of the message, not at the message start, so it does not feel like a useful “jump to this pinned thing”.

Current code uses:

```js
row.scrollIntoView({ block: 'center', behavior: 'smooth' });
```

Local patch that felt right for this interaction:

```js
row.scrollIntoView({ block: 'start', inline: 'nearest', behavior: 'smooth' });
```

Verification after the local patch: clicking a pinned popover item closed the popover and moved the target row’s top to the top of `.messages-shell` (`deltaTop: 0` in the browser measurement), instead of centering the long message.

I would fix both before merge. They are small, extension-local changes and do not alter the storage/security shape.

## Out of scope: selected-text pins

One related product idea came up during testing: for very long final answers, pinning the whole message row may still be too coarse. A future version might let the user select a specific paragraph/sentence/code block and pin that excerpt, so the pinned list can jump back to the exact part of a long answer.

I am intentionally not asking for that in this PR. It changes the product model from “pin this message row” to “pin this text range inside a rendered answer”, which brings separate design and engineering questions: how to persist a text-range anchor across Markdown re-renders and reloads, how to handle edited/regenerated/virtualized assistant content, what preview/removal should look like in the popover, and how to avoid interfering with normal text selection/copy behavior.

So my recommendation for this PR is narrower: make the existing message-level pins reliable and predictable first. If selected-text pins still feels valuable after this lands, it should be split into a separate issue/RFC rather than folded into this review.

nesquena-hermes:
Thanks again @franksong2702 — both round-2 issues are now fixed on the branch (latest head), and the storage/security shape is unchanged.

## 1. Inline pin/unpin reliability
Applied your diagnosed fix: the inline `.hwx-pin-btn` is now a **stable 24×24 button** with `position:relative; z-index:2; pointer-events:auto`, and `svg { pointer-events:none }` so `elementFromPoint()` always resolves to the BUTTON instead of `.msg-foot` or the SVG child. The active state keeps `padding:0` so the hit target doesn't shift between pin and unpin (your `23x17 → 21x17` observation). This is exactly the patch you verified locally (5 toggles in a row).

## 2. Popover click-to-jump centering long messages
Changed `scrollIntoView({ block:'center' })` → `{ block:'start', inline:'nearest' }`, so a pinned entry jumps the row's **top** to the top of the messages shell rather than landing mid-message.

## Out-of-scope (agreed)
Selected-text / excerpt pins — left out of this PR as you suggested; it's a separate product model (persisting a text-range anchor across re-renders) and belongs in its own issue/RFC.

Both changes are extension-local CSS/JS. Gates green (validate / safety-scan / validator self-test).

franksong2702:
Follow-up validation on latest head (`3de2e8c0...`): this version looks good from my side.

I tested the refreshed #19 build on the local 8787 runtime in both:

- desktop browser
- iPad PWA mode

The two round-2 interaction fixes both passed in real use:

- inline pin / unpin is responsive now; the click target no longer feels dead or unreliable
- clicking a pinned entry from the header popover now jumps to the start of the pinned message instead of landing in the middle of a long answer

I do not see a remaining UX blocker for this extension. GitHub currently shows the PR as conflicting with current `main`, so my recommendation is: rebase / resolve the merge conflict, rerun the gates, and then this is good to merge.

PR #23, opened 2026-06-28
Add model-favorites extension (star + promote favorite models)

nesquena-hermes:
## Summary

Adds the **`model-favorites`** extension — star your most-used models in the composer model picker. Favorited models are promoted to a **★ Favorites** group pinned to the top of the dropdown for one-click switching.

- A star toggle on every model row (shown on hover, always visible once favorited).
- **Provider-aware**: the same model id under two providers is two distinct favorites.
- Persists locally (`localStorage`); re-applies after the picker re-renders (open / search / select) via a `MutationObserver`.
- Pure DOM-injection over the existing `#composerModelDropdown` `.model-opt` rows — **no core changes, no backend**.

## Credit

The provider-aware favorite design follows closed core PR **nesquena/hermes-webui#3578** (@starship-s). That was a full core re-implementation against the pre-grouping picker; this is a lighter client-side extension over the current grouped picker. (This is the model-favorites idea Nathan flagged — a fuller "favorites in the footer provider window" version may extend it later.)

Tested by seeding multiple models into the picker and driving it via headless Chrome/CDP:

- ✅ stars decorate every model row (4 models → 4 stars)
- ✅ starring 2 models builds a **★ Favorites** group with exactly those 2
- ✅ **provider-aware** keying (same id under different providers tracked separately)
- ✅ selecting a favorite sets `modelSelect.value` to that model
- ✅ un-favorite works from both the row and the Favorites group
- ✅ **survives re-render** — fav group + stars re-applied, no duplicate groups
- ✅ **persists across full reload**
- ✅ polish pass: star sits inline (right edge) with the model name, matching the main rows — visually confirmed

- `webui_api.read/write: []` — no `fetch`/`/api/*`; selecting a model calls the existing `window.selectModelFromDropdown(...)`
- `network_external: false`; no eval; no cookies
- `storage.owned: ["hermes-ext-model-favorites"]` — the one key written
- `dom.owned: true`, `dom.mutates_core_views: true` — injects stars + the Favorites group into the dropdown; all rendered text is escaped

## Known limitation

Relies on the current model-picker DOM (`.model-opt` / `.model-opt-id` / `.model-opt-name` / `.model-opt-provider`); a core rename would need an update — standard for a DOM-injection extension.

### DIFF
--- a/extensions/README.md
+++ b/extensions/README.md
@@ -16,10 +16,11 @@ Do not add shared runtime code here unless multiple accepted extensions already
 need it and maintainers agree on the shared contract.
 
 ## Entries
-
 - `desktop-companion`: trusted local Desktop Companion entry and first
   sidecar-class extension candidate.
 - `mobile-conversations`: phone-only floating Conversations button, same-location
   drawer close X, and long-press shortcuts for the existing Hermes WebUI mobile drawer.
 - `message-pins`: pin individual messages in a conversation, with a header
   popover, click-to-jump, and client-side per-session persistence.
+- `model-favorites`: star your most-used models; favorites are promoted to a
+  ★ Favorites group at the top of the composer model picker.
--- a/extensions/model-favorites/README.md
+++ b/extensions/model-favorites/README.md
@@ -1,0 +1,106 @@
+# Model Favorites
+
+Model Favorites is a trusted local Hermes WebUI extension that lets you star the
+models you use most in the composer model picker. Favorited models are promoted
+to a **★ Favorites** group at the top of the dropdown for one-click switching.
+
+## What It Does
+
+- Adds a star toggle to each model row in the composer model picker (visible on
+  hover, and always visible once a model is favorited).
+- Promotes favorited models into a **★ Favorites** group pinned to the top of
+  the dropdown.
+- **Provider-aware**: the same model id offered by two providers is two distinct
+  favorites, so you can favorite exactly the one you mean.
+- Persists favorites locally (`localStorage`), so they survive reloads.
+- Re-applies after the picker re-renders (open / search / select) via a
+  `MutationObserver`.
+
+## Current Shape
+
+```text
+Hermes WebUI page
+  -> manifest-bundled extension assets
+  -> /extensions/assets/model-favorites.js + .css
+  -> decorates #composerModelDropdown .model-opt rows with a star
+  -> injects a "★ Favorites" group at the top
+  -> localStorage: hermes-ext-model-favorites
+```
+
+This extension is `static-ui` / manifest-bundle only. It does not add backend
+routes, start a sidecar, access external networks, read or write files, or use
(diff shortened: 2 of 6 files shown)


nesquena-hermes:

franksong2702 (approved):
Smoke-tested this on a clean local 8787 runtime from current `master` (`v0.51.757`), using a temporary aggregate extension bundle that preserved the already-installed extensions and added this PR head (`687b60930a23`).

What I verified:

- `node --check extensions/model-favorites/assets/model-favorites.js` passes.
- Manifest/permission shape matches the behavior: no WebUI API access, no sidecar/native host/network, DOM mutation declared, and storage scoped to `hermes-ext-model-favorites`.
- Runtime `/api/extensions/status` loaded the extension with no warnings; JS and CSS assets both served with HTTP 200.
- Opening the composer model dropdown decorated the real model rows with favorite stars.
- In my local dropdown, 80 model rows were decorated.
- Favoriting `deepseek-v4-flash` wrote the expected localStorage record and injected the `Favorites` group at the top of the dropdown.
- I restored the previous `hermes-ext-model-favorites` localStorage value after the smoke test.

Looks good from my side.

nesquena-hermes:

This PR has been through a full QA pass — an independent **Codex** security+breakage gate (4 rounds) and a hands-on browser pass on a real model picker. Everything caught has been fixed and re-verified gate-clean. Summary of what was addressed:

- **Codex (BREAK):** the MutationObserver could self-trigger (the `applying` guard was reset synchronously while the observer callback fires on a later microtask, and `decorateRows()` rewrote `star.innerHTML` every pass) → rAF/decorate loop risk. **Fixed** — disconnect the observer during `apply()` and reconnect after, and only rewrite the star DOM when the filled state actually changed.
- **Codex (BREAK):** provider-awareness was scraped from the visible `.model-opt-provider` chip, which core **omits** under provider-grouped layouts; the real `providerId` lives only in core's row `onclick` closure. Duplicate model ids across providers could select the wrong provider. **Fixed** over two rounds: derive the provider from the enclosing core group wrapper `.model-group-body[data-group]`, **ignoring nested `.model-group-body.sub` vendor subgroups** (whose `data-group` is `provider::vendor`, not a real provider id); selecting a favorite re-clicks the live core row matched on **id + provider** to inherit its exact selection, falling back to `selectModelFromDropdown(id, provider)` only when no live row exists.

**Verification performed:**
- Functional E2E in a real seeded conversation on an isolated server — stars decorate every row, the "★ Favorites" group builds at the top and survives a dropdown re-render, star clicks do NOT trigger model selection (stopPropagation), provider-aware persistence, clean visuals (vision-checked)
- Codex security+breakage gate: **SAFE TO SHIP** (re-gated to confirm each finding resolved — it took 4 rounds, each surfacing a deeper DOM-contract nuance)

PR #24, opened 2026-06-28
Add custom-avatar extension (custom assistant avatar image)

nesquena-hermes:
## Summary

Adds the **`custom-avatar`** extension — give the assistant a custom avatar image in the chat transcript. The assistant's role badge is normally a single-letter glyph; this swaps in an image of your choice.

- Click any assistant avatar badge (`.role-icon.assistant`) → a small picker to upload or clear an image.
- The image is **downscaled to a 64×64 square** (in-browser `<canvas>`) and stored as a small data-URL in `localStorage`.
- Re-applied to every assistant turn after the transcript re-renders (streaming/scroll/reload) via a `MutationObserver`; falls back to the letter glyph when cleared.
- Pure DOM-injection, image processed entirely in-browser (**never uploaded**) — no core changes, no backend.

## Scope note + why no core bug was filed

This customizes the **assistant** avatar because that's the avatar element that exists. The WebUI deliberately renders **no avatar on user messages** (right-aligned bubble; position identifies the sender), so there's no user-side avatar slot.

The backlog flagged a possible core bug from closed issue #2677 ("assistant shows the user's avatar"). I checked empirically: **on current master there are no avatar images at all** — both roles render letter glyphs — so that bug does not reproduce, and I did **not** file a spurious core issue. This extension adds the personalization layer instead.

Tested against real assistant turns (mock provider) via headless Chrome/CDP:

- ✅ extension loads; assistant role badge present after a turn
- ✅ setting an avatar replaces the glyph with an `` (glyph hidden, set-class added)
- ✅ a **new assistant turn also gets the avatar** (MutationObserver re-apply)
- ✅ **persists across full reload** and re-applies to all assistant turns
- ✅ **clear** removes it from storage and **restores the letter glyph** ("H")
- ✅ picker UI opens cleanly under the badge ("Assistant avatar" + Upload button) — visually confirmed

- `webui_api.read/write: []` — no `fetch`/`/api/*`
- `network_external: false` — the image never leaves the browser (FileReader + canvas only)
- `storage.owned: ["hermes-ext-assistant-avatar"]` — the one key written (a downscaled data-URL)
- `dom.owned: true`, `dom.mutates_core_views: true` — injects an `` into the assistant badge + a picker popover
- only validated `data:image/(png|jpeg|gif|webp);base64,…` values are ever applied

## Known limitations

- Assistant-only (the user role has no avatar slot by design).
- Per-browser (`localStorage`), not synced across devices.
- Relies on the `.role-icon.assistant` badge; a core rename would need an update.

### DIFF
--- a/extensions/README.md
+++ b/extensions/README.md
@@ -16,10 +16,11 @@ Do not add shared runtime code here unless multiple accepted extensions already
 need it and maintainers agree on the shared contract.
 
 ## Entries
-
 - `desktop-companion`: trusted local Desktop Companion entry and first
   sidecar-class extension candidate.
 - `mobile-conversations`: phone-only floating Conversations button, same-location
   drawer close X, and long-press shortcuts for the existing Hermes WebUI mobile drawer.
 - `message-pins`: pin individual messages in a conversation, with a header
   popover, click-to-jump, and client-side per-session persistence.
+- `custom-avatar`: give the assistant a custom avatar image in the chat
+  transcript (downscaled + stored locally; assistant-only).
--- a/extensions/custom-avatar/README.md
+++ b/extensions/custom-avatar/README.md
@@ -1,0 +1,125 @@
+# Custom Assistant Avatar
+
+Custom Assistant Avatar is a trusted local Hermes WebUI extension that lets you
+give the assistant a custom avatar image in the chat transcript. The assistant's
+role badge is normally a single-letter glyph; this swaps in an image of your
+choice.
+
+## What It Does
+
+- Click any assistant avatar badge (`.role-icon.assistant`) to open a small
+  picker: upload an image, or remove the current one.
+- The image is **downscaled to a 64×64 square and stored locally** as a data-URL
+  (`localStorage`), so it stays small and survives reloads.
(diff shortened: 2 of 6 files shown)


nesquena-hermes:

franksong2702 (approved):
Smoke-tested this on a clean local 8787 runtime from current `master` (`v0.51.757`), using a temporary aggregate extension bundle that preserved the already-installed extensions and added this PR head (`cf2267a22699`).

What I verified:

- `node --check extensions/custom-avatar/assets/custom-avatar.js` passes.
- Manifest/permission shape matches the behavior: no WebUI API access, no sidecar/native host/network, DOM mutation declared, and storage scoped to `hermes-ext-assistant-avatar`.
- Runtime `/api/extensions/status` loaded the extension with no warnings; JS and CSS assets both served with HTTP 200.
- In a real existing conversation, the extension found 4 assistant role icons.
- Clicking an assistant icon opened the avatar picker.
- Setting a tiny test PNG data URL replaced all 4 assistant icons with avatar images.
- Clearing the avatar restored the glyph state; image count returned to 0.
- I restored the previous `hermes-ext-assistant-avatar` localStorage value after the smoke test.

Looks good from my side.

PR #25, opened 2026-06-28
Add external-app-tab extension (pin a compatible self-hosted web app via iframe)

nesquena-hermes:
## Summary

Adds the **`external-app-tab`** extension — pin a compatible self-hosted web app (Grafana, Vaultwarden, Linkwarden, a personal dashboard) as a tab inside the WebUI via an `<iframe>`. Adds a left-rail button that opens a full-area overlay framing a URL you configure.

- Rail button → full-area overlay with a top bar (title, **Configure**, **Open ↗**, **Close**).
- **Configure** dialog: label + `http(s)` URL, stored locally; empty-state prompt until set.
- URL validation (http/https only — rejects `javascript:` etc.); config persists in `localStorage`.
- No backend, **no fetches of its own** — it only sets an `<iframe src>`, loaded by the browser under the page CSP.

## ⚠️ Dependency

To frame an **external** origin, the operator must allow it via the core knob **nesquena/hermes-webui#5091** (`HERMES_WEBUI_CSP_FRAME_EXTRA`):

```bash
export HERMES_WEBUI_CSP_FRAME_EXTRA="https://your-app.example.com"
```

Same-origin / loopback-reverse-proxied URLs need no core change. If a frame is blocked by CSP, the overlay shows a hint with the exact env value to set.

## End-to-end verification (live browser, against the #5091 capability)

- ✅ **the core CSP knob works**: the served header carried `frame-src 'self' https://example.com` (enforced + report-only) when `HERMES_WEBUI_CSP_FRAME_EXTRA` was set
- ✅ rail button appears and opens the overlay
- ✅ empty-state prompt + config dialog render cleanly (visually confirmed)
- ✅ configuring a URL creates the iframe with the correct `src` and updates the title
- ✅ URL validation rejects `javascript:`/non-http; config **persists across reload**
- ✅ Open ↗ / Close / Escape behave; repo gates green (validate / safety-scan / registry / self-test)

**Honest test-environment caveat:** whether a *given remote app actually paints* depends on two things beyond this extension — (1) the operator allowing the origin via `HERMES_WEBUI_CSP_FRAME_EXTRA` (verified working), and (2) the **target app's own** `X-Frame-Options` / `frame-ancestors` permitting embedding (the remote site's choice). The WebUI itself sends `X-Frame-Options: DENY`, so it correctly can't be framed by anyone — including this extension. The isolated test box has no outbound network to a real framable site, so I verified the entire WebUI-side chain (CSP header, iframe creation, src, config, validation) rather than a live remote render. Both realities are documented in the README's Known Limitations.

## Permissions disclosure

- `network_external: true` — the whole point is embedding a user-configured external URL in an iframe (subject to CSP); the extension makes no `fetch()` of its own
- `storage.owned: ["hermes-ext-external-app"]`; `dom.owned: true`/`mutates_core_views: true` (rail button + overlay)
- no WebUI API calls, no cookies, no sidecar/native host; user strings escaped where rendered

### DIFF
--- a/extensions/README.md
+++ b/extensions/README.md
@@ -39,3 +39,6 @@ need it and maintainers agree on the shared contract.
   Appearance picker via the theme-registration API.
 - `mobile-haptics`: short device vibration when an assistant turn finishes
   (Android / Android-PWA; no-ops on desktop and iOS).
+- `external-app-tab`: pin a self-hosted web app (Grafana, Vaultwarden, a
+  dashboard) as a tab inside the WebUI via an iframe (needs the core
+  `HERMES_WEBUI_CSP_FRAME_EXTRA` knob for external origins).
--- a/extensions/README.md
+++ b/extensions/README.md
@@ -39,6 +39,6 @@ need it and maintainers agree on the shared contract.
   Appearance picker via the theme-registration API.
 - `mobile-haptics`: short device vibration when an assistant turn finishes
   (Android / Android-PWA; no-ops on desktop and iOS).
-- `external-app-tab`: pin a self-hosted web app (Grafana, Vaultwarden, a
-  dashboard) as a tab inside the WebUI via an iframe (needs the core
-  `HERMES_WEBUI_CSP_FRAME_EXTRA` knob for external origins).
+- `external-app-tab`: pin a compatible self-hosted web app (Grafana, a
+  dashboard, or another framable tool) as a tab inside the WebUI via an iframe
+  (needs the core `HERMES_WEBUI_CSP_FRAME_EXTRA` knob for external origins).
--- a/extensions/external-app-tab/README.md
+++ b/extensions/external-app-tab/README.md
@@ -91,9 +91,18 @@ This is trusted local code. Current disclosed behavior:
 - does NOT access cookies, loopback sidecars, the filesystem, or native hosts
 - does NOT itself `fetch()` any remote resource — it only sets an iframe `src`
 
-The embedded app runs in its own browsing context (a separate origin in its own
-iframe); it cannot read the WebUI's DOM or session. All user-supplied strings
-(label, URL) are escaped where rendered.
+The embedded app runs in its own `<iframe>` with an explicit `sandbox`
+allow-list: `allow-scripts allow-forms allow-popups
+allow-popups-to-escape-sandbox allow-same-origin allow-downloads`. This lets a
+real web app function (run its scripts, submit forms, open links/downloads)
+while still being a sandboxed frame. **Honest tradeoff:** a functional app
+generally needs both `allow-scripts` and `allow-same-origin`, and that pairing
+relaxes the sandbox's origin barrier — so for a **same-origin** target the frame
+is effectively trusted browser content, not strongly isolated. Treat any app you
+embed as trusted: you opt in per-origin by allow-listing it via
+`HERMES_WEBUI_CSP_FRAME_EXTRA`, and cross-origin targets remain bounded by the
+same-origin policy. All user-supplied strings (label, URL) are escaped where
+rendered.
 
 ## Compatibility
 
--- a/extensions/external-app-tab/README.md
+++ b/extensions/external-app-tab/README.md
@@ -1,9 +1,9 @@
 # External App Tab
 
 External App Tab is a trusted local Hermes WebUI extension that pins a
-self-hosted web app — Grafana, Vaultwarden, Linkwarden, a personal dashboard —
-as a tab inside the WebUI via an `<iframe>`. It adds a button to the left rail
-that opens a full-area panel framing a URL you configure.
+compatible self-hosted web app — Grafana, Vaultwarden, Linkwarden, a personal
+dashboard — as a tab inside the WebUI via an `<iframe>`. It adds a button to
+the left rail that opens a full-area panel framing a URL you configure.
 
 It works best for **framable dashboards, status pages, and local tools** that
 don't require a cross-origin cookie login. It provides an iframe slot — not a
--- a/extensions/external-app-tab/README.md
+++ b/extensions/external-app-tab/README.md
@@ -1,9 +1,14 @@
 # External App Tab
 
-External App Tab is a trusted local Hermes WebUI extension that pins any
+External App Tab is a trusted local Hermes WebUI extension that pins a
 self-hosted web app — Grafana, Vaultwarden, Linkwarden, a personal dashboard —
 as a tab inside the WebUI via an `<iframe>`. It adds a button to the left rail
 that opens a full-area panel framing a URL you configure.
+
+It works best for **framable dashboards, status pages, and local tools** that
+don't require a cross-origin cookie login. It provides an iframe slot — not a
+guarantee that any arbitrary app will work correctly embedded (see **Known
+Limitations** for the third-party-iframe auth/cookie caveats).
 
 ## What It Does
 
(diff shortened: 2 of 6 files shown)


nesquena-hermes:

franksong2702:
One security/model mismatch I would fix before merge:

In `extensions/external-app-tab/assets/external-app-tab.js`, the iframe setup says:

```js
// Sandboxed but functional; allow the framed app its own scripts/forms.
frame.setAttribute('referrerpolicy', 'no-referrer');
```

but the code never sets a `sandbox` attribute on the iframe. So the current implementation is an **unsandboxed** user-configured iframe, constrained by CSP / browser same-origin rules, but not by iframe sandboxing.

That distinction matters for this extension's threat model:

- cross-origin apps are still limited by SOP, but they are not sandboxed;
- same-origin or reverse-proxied apps are especially sensitive, because an unsandboxed same-origin frame can interact with the parent origin much more than a sandboxed frame could;
- the PR/README currently gives reviewers the impression that there is an iframe sandbox boundary when there is not.

I think the fix should be one of these:

1. actually set a `sandbox` attribute and document the exact allow-list/tradeoffs (`allow-scripts`, `allow-forms`, `allow-popups`, etc.), or
2. intentionally keep it unsandboxed for compatibility, but remove the “sandboxed” claim and make the permissions/README explicit that the embedded app is trusted browser content once the operator allows it via `HERMES_WEBUI_CSP_FRAME_EXTRA`.

The CSP allowlist piece in #5091 is still useful either way; this is specifically about aligning the iframe sandbox claim with the implementation.

nesquena-hermes:
Thanks @franksong2702 — good catch on the sandbox claim mismatch; fixed by actually sandboxing the frame (your option 1).

The iframe now sets an explicit `sandbox` allow-list:
```
allow-scripts allow-forms allow-popups allow-popups-to-escape-sandbox allow-same-origin allow-downloads
```
so an embedded app can function while being a genuinely sandboxed frame. The misleading "sandboxed but functional" comment is replaced with one that states the **honest tradeoff**: a functional web app generally needs `allow-scripts` + `allow-same-origin`, and that pairing relaxes the sandbox's origin barrier for a same-origin target — so an embedded app is trusted browser content, opted-in per-origin via `HERMES_WEBUI_CSP_FRAME_EXTRA`. The README's Trust & Permissions section now documents the exact allow-list and that tradeoff explicitly. Gates green.

nesquena-hermes:

- **@franksong2702 (earlier round):** the iframe was described as "sandboxed" but no `sandbox` attribute was set. **Fixed** — the frame now sets an explicit `sandbox` allow-list + `referrerpolicy="no-referrer"`, and the README documents the `allow-scripts`+`allow-same-origin` same-origin tradeoff honestly.

**Verification performed:**
- Functional E2E in a real seeded conversation on an isolated server (not a synthetic harness) — drove the configure flow, verified the `sandbox` attribute is actually set on the live iframe, confirmed no XSS via the user-supplied URL/label (title via `textContent`), and that a CSP-blocked frame degrades gracefully with the operator hint

rodboev (changes requested):
I reran this on current core from `origin/master` with your PR head loaded locally.

Frank's earlier sandbox issue still reproduces in the live DOM. After configuring a same-origin URL and opening the tab, the rendered iframe had:

- `src = "http://127.0.0.1:8792/"`
- `referrerpolicy = "no-referrer"`
- `sandbox = null`

The PR head source still matches that result in `extensions/external-app-tab/assets/external-app-tab.js`: it sets `referrerpolicy`, but I still don't see a `frame.setAttribute('sandbox', ...)` call in this branch.

So the branch still looks unsandboxed while the review thread says this was fixed.

nesquena-hermes:
## The iframe IS sandboxed at the PR head — verified in the live DOM 🙏 @rodboev @franksong2702

Thanks Rod for re-running this. I dug in, and the `sandbox = null` result is a stale-asset artifact in the test bundle, not the branch state — the sandbox fix is present and active at the PR head.

**At the live head (`ccddae1`, rebased clean onto `main`):**
```
$ grep -n "setAttribute('sandbox'" extensions/external-app-tab/assets/external-app-tab.js
142:    frame.setAttribute('sandbox', 'allow-scripts allow-forms allow-popups allow-popups-to-escape-sandbox allow-same-origin allow-downloads');
143:    frame.setAttribute('referrerpolicy', 'no-referrer');
```
This landed in commit `5f6a821` (2026-06-29), before the review — so any bundle showing `sandbox=null` was serving a pre-fix copy of the asset (the temporary aggregate bundle / `?v=` asset cache carrying the old version).

**Reproduced your exact setup in a live browser** (single extension served from its own worktree on an isolated port, cache-busted page): configured a same-origin URL → opened the tab → read the rendered iframe attributes from the live DOM via CDP:
```json
{
  "src": "http://127.0.0.1:8789/health",
  "sandbox": "allow-scripts allow-forms allow-popups allow-popups-to-escape-sandbox allow-same-origin allow-downloads",
  "referrerpolicy": "no-referrer",
  "sandboxIsNull": false
}
```
So the live frame is sandboxed with the explicit allow-list, exactly as the comment/README describe (including the honest `allow-scripts` + `allow-same-origin` tradeoff note for same-origin embeds).

### Codex security/breakage gate
Also ran the extension Codex gate against the head: **SAFE TO SHIP**. It independently verified the sandbox + `referrerpolicy`, that `src` only accepts absolute `http:`/`https:` URLs (`javascript:`/`data:`/relative rejected), label rendering via `textContent`/`escapeHtml`, no rail-button collision against core's `.nav-tab[data-panel]` ordering, and that the disclosed permissions (`network_external`, `embeds_iframe`, owned localStorage key) match the code.

Could you re-pull the head (`ccddae1`) into a fresh bundle and confirm? It's rebased onto `main` (0 behind), validators + safety scan green. Tagging you both for the final look.

franksong2702:
One more product-boundary issue before merge: the README/PR needs to be clearer that this extension only provides an iframe slot, not a guarantee that any arbitrary self-hosted app will work correctly inside that iframe.

Hands-on validation with a local Paperclip instance exposed the common failure mode:

- after adding the Paperclip origin to `HERMES_WEBUI_CSP_FRAME_EXTRA`, the app can be reached in the iframe
- Paperclip's login request succeeds (`POST /api/auth/sign-in/email` returns 200) and the app creates a server-side session
- the next iframe request to `GET /api/auth/get-session` returns 401 because the browser did not send the Paperclip session cookie from inside the embedded cross-origin context

That is not something this extension can fix with frontend JS. It is the normal browser/security boundary for authenticated cross-origin iframes: third-party cookie blocking, `SameSite`, `Secure`, HTTPS requirements, and the target app's own frame/cookie policy all matter. `HERMES_WEBUI_CSP_FRAME_EXTRA` only answers "may Hermes embed this origin?" It does not make the target app iframe-compatible, and it does not make login/session cookies work in a third-party iframe.

Please update the README / PR description before merge to set that expectation explicitly. I would like the docs to say something like:

- works best for framable dashboards, status pages, local tools, and apps that do not require cross-origin cookie login
- authenticated apps may fail inside the iframe even when the origin is allow-listed, because their session cookies may not be available in a third-party iframe
- if login succeeds but the embedded app remains logged out / returns 401, use **Open in new tab** or configure the target app for iframe use
- stable embedded login usually requires a same-origin reverse proxy, or HTTPS plus appropriate cookie settings such as `SameSite=None; Secure`, and the target app must allow framing via its own `frame-ancestors` / `X-Frame-Options` policy
- `HERMES_WEBUI_CSP_FRAME_EXTRA` is necessary for external origins, but not sufficient for all apps

This is not a code-level blocker for the current iframe implementation, but it is a merge blocker for the product claim. The current "pin any self-hosted web app" wording is too broad unless the limitations for login/session-heavy apps are spelled out clearly.

nesquena-hermes:
## Docs updated for the cross-origin iframe auth boundary 🙏 @franksong2702

Great catch on the Paperclip session-cookie failure mode — that's exactly the third-party-iframe boundary and not something the extension can fix in JS. Updated the README to set the expectation explicitly (rebased onto current `main`, no code change):

**Intro** — softened "pin *any* self-hosted web app" → "pin a self-hosted web app … works best for framable dashboards, status pages, and local tools that don't require a cross-origin cookie login. It provides an iframe slot — not a guarantee that any arbitrary app will work embedded."

**Known Limitations** — added your points, near-verbatim:
- `HERMES_WEBUI_CSP_FRAME_EXTRA` is **necessary but not sufficient** — it answers "may Hermes embed this origin?", not "will the app work embedded?"
- Authenticated apps may fail even when allow-listed, because session cookies may not be sent from a third-party iframe context (3p-cookie blocking, `SameSite`, `Secure`/HTTPS, the app's own frame/cookie policy). Common tell: login 200s but the next in-frame request 401s.
- If login succeeds but the app stays logged out / 401 → use **Open in new tab**, or configure the app for iframe use.
- Stable embedded login usually needs a same-origin reverse proxy, or HTTPS + `SameSite=None; Secure` + a permissive framing policy.

Since this was a caught review item, not self-merging — @franksong2702 @rodboev, one more look at the README wording when you get a moment. The iframe sandbox + code are unchanged from the prior review-clean state (Codex SAFE).

PR #27, opened 2026-06-28
Add session-export-pdf extension (export conversation to PDF / Markdown)

nesquena-hermes:
## Summary

Adds the **`session-export-pdf`** extension — export the current conversation to a clean **PDF** (via the browser print dialog with a print-styled layout) or **copy it as Markdown**. Adds a download button to the app titlebar.

- Titlebar button (next to Reload), shown only when a conversation is open → menu with **Export to PDF** + **Copy as Markdown**.
- PDF path clones the rendered transcript into an off-screen print container and `window.print()`s with a scoped `@media print` stylesheet so the output is just the titled transcript (roles, code blocks, links preserved), not the app chrome.
- **No backend, no bundled PDF library, no network.**

## Credit
Design reference: closed core PR **nesquena/hermes-webui#3425** (@vanshaj-pahwa) — the print-markup/formatting approach. Routed to extensions to keep core lean.

## Full pre-push QA (fresh-port sim server, 12-message conversation)
Per the new `webui-extension-qa` process (the one that came out of @franksong2702's #19 review):
- ✅ titlebar button placed cleanly between New-Chat and Reload — **no overlap, vision-verified**
- ✅ menu opens with both items; button hides when no conversation is open
- ✅ **PDF print root builds with all 12 messages**, title, and **preserved code blocks**; `window.print()` fires and the `afterprint` handler cleans up the print root + body class
- ✅ **`@media print` isolation rule verified served** (hide app chrome / show only the print root / `@page` margins)
- ✅ **Copy as Markdown** produces a full titled doc (You/Assistant sections, code content)
- ✅ gates green (validate / safety-scan / registry / validator self-test)

## Permissions
- `network_external:false`; no API calls; no storage; `uses_print` + `uses_clipboard`; `dom.owned:true` (titlebar button + temporary print container, removed after print). Print body reuses core's already-sanitized rendered HTML (cloned, not re-parsed).

### DIFF
--- a/extensions/README.md
+++ b/extensions/README.md
@@ -28,3 +28,5 @@ need it and maintainers agree on the shared contract.
   transcript (downscaled + stored locally; assistant-only).
 - `theme-creator`: build your own theme with a live color editor; custom themes
   register into the native Appearance picker (needs core theme-registration).
+- `session-export-pdf`: export the current conversation to PDF (print) or copy
+  it as Markdown, via a titlebar button.
--- a/extensions/session-export-pdf/README.md
+++ b/extensions/session-export-pdf/README.md
@@ -1,0 +1,121 @@
+# Session Export to PDF
+
+Session Export to PDF is a trusted local Hermes WebUI extension that lets you
+export the current conversation to a clean **PDF** (via the browser's print
+dialog, with a print-styled layout) or **copy it as Markdown**. It adds an export
+button to the app titlebar.
+
+## What It Does
+
+- Adds an export button to the app titlebar (next to Reload); it shows only when
+  a conversation is open.
+- Clicking it opens a small menu:
+  - **Export to PDF** — clones the rendered transcript into a print-styled,
+    off-screen container and calls `window.print()` with a scoped `@media print`
+    stylesheet, so the printed/saved PDF is just the conversation (titled, with
+    roles, code blocks, links preserved) rather than a raw `Ctrl+P` of the whole
+    app chrome. Use your browser's "Save as PDF" destination.
+  - **Copy as Markdown** — copies the conversation as `# title` + `## You` /
+    `## Assistant` sections to the clipboard.
+- No backend, **no bundled PDF library**, no network. Everything is done in the
+  browser from the already-rendered transcript.
+
+## Credit
+
+Design reference: closed core PR
+[#3425](https://github.com/nesquena/hermes-webui/pull/3425) (@vanshaj-pahwa),
+which added a print-markup/formatting export to core. This is the extension form
+of that idea (routed to extensions to keep core lean).
+
+## Current Shape
+
+```text
+Hermes WebUI page
+  -> manifest-bundled extension assets
+  -> /extensions/assets/session-export-pdf.js + .css
+  -> titlebar button -> menu -> clone #messages rows into #hwxPrintRoot
+     + @media print { show only #hwxPrintRoot } -> window.print()
+  -> (or) copy transcript as Markdown to the clipboard
+```
+
+This extension is `static-ui` / manifest-bundle only. It does not add backend
+routes, start a sidecar, access external networks, read/write files, or use
+native host APIs.
+
+## Capabilities
+
+- `manifest-bundle`
+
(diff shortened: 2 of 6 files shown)


nesquena-hermes:
@franksong2702 @rodboev — review welcome when you have a moment, both **code** and a **hands-on browser pass**.

This one was built and QA'd with the full pre-push process that came out of @franksong2702's #19 review: a fresh-port simulated server with a real long conversation, exercised end-to-end, screenshots vision-checked for positioning / overlap / styling, and the safety-critical behavior verified directly. The PR description's "Full pre-push QA" section lists exactly what was driven and checked.

If anything feels off in real use — positioning collisions with core controls, styling, or behavior the automated gates wouldn't catch — please flag it. Gates green, rebased on current main.

franksong2702:
I would clarify or fix the export scope before merge.

`collectRows()` reads only the currently rendered DOM:

```js
const container = $('messages');
container.querySelectorAll('[data-msg-idx]').forEach(...)
```

and it also skips rows that are hidden/collapsed or not currently laid out. That means this extension exports the **loaded/rendered transcript window**, not necessarily the full conversation/session.

For short QA sessions that is fine, but in real long Hermes conversations the message pane can be paginated/windowed. Older messages may not be present in `#messages` at export time, so `Export conversation` / `Copy as Markdown` can silently omit part of the conversation while still producing a clean-looking document.

I see two reasonable paths:

1. keep the no-API extension shape, but rename/document the behavior as “export loaded transcript” or show a warning/count that makes the scope explicit; or
2. if the intent is truly “full conversation”, read the complete session through a WebUI API and declare that permission, then build the export from durable session data rather than the current DOM.

I would not ship it as “export current conversation” while the implementation is DOM-window based, because users will assume the export is complete.

nesquena-hermes:
Thanks @franksong2702 — fixed, keeping the no-API shape (your option 1) and making the scope honest so a windowed export is never silently presented as complete.

Added `transcriptMaybeWindowed()` — detects a `[data-virtual-spacer]` node or `window._virtualizeTranscript === true`. When the transcript is windowed/virtualized:
- the **PDF** prints a visible note in the header: *"this export covers the currently-loaded transcript … scroll to the top to load earlier messages for a complete copy"* and a toast fires on export;
- the **Markdown** copy gets an equivalent blockquote note and the toast says "loaded transcript … scroll up for earlier messages."

The README's Known Limitations now states plainly that scope = the **loaded/rendered transcript, not guaranteed the full session**, with the scroll-to-top (or disable-virtualization) workaround for a complete copy. For the common default case (virtualization off, full transcript rendered) there's no note and it's the complete conversation. Gates green.

nesquena-hermes:

- **Codex (SECURITY):** the PDF print path reparsed message-body `innerHTML`. **Fixed** — the print DOM is now built with `createElement`/`textContent` and a **sanitized `cloneNode`** of each `.msg-body` (strips `on*` handlers, script/style/iframe, off-origin `src`/`poster`/`srcset`, `javascript:` hrefs).
- **Codex (SECURITY):** off-origin `` in rendered markdown would fire cross-origin requests on export while `network_external:false` is declared. **Fixed** — `sanitizeClone()` strips off-origin media (same-origin + `data:` preserved), keeping the no-network disclosure honest.
- **Codex (DISCLOSURE):** windowed-transcript detection missed core's server-paginated load-earlier state. **Fixed** — now also detects `#loadOlderIndicator` / `.message-window-load-earlier`.

**Verification performed:**
- Functional E2E in a real seeded conversation on an isolated server (not a synthetic harness) — drove markdown + print export, verified off-origin images are stripped (same-origin kept), the windowed note appears only when the transcript is actually partial, and hidden anchor/worklog nodes are excluded

franksong2702:
Validated locally on the clean 8787 extension bundle. The session titlebar export button appears, and clicking it opens the `Export to PDF` / `Copy as Markdown` menu; no extension-level console errors in the smoke run.

This version looks good from my side. The PR is currently merge-conflicting with `main`, so please rebase/update the branch; once conflicts are resolved and checks stay green, this is mergeable.

PR #29, opened 2026-06-28
Add voicevox-tts extension (local VOICEVOX as a TTS engine)

nesquena-hermes:
## Summary

Adds the **`voicevox-tts`** extension — use **VOICEVOX** (a free, locally-hosted Japanese TTS engine) as a TTS engine in Hermes WebUI. Once selected in **Settings → TTS Engine**, both the per-message **Listen** button and hands-free **voice mode** synthesize speech through your local VOICEVOX server.

- Registers a **VOICEVOX (local)** engine via the core `window.registerHermesTtsEngine` hook.
- Synthesizes via the standard VOICEVOX two-step flow: `POST /audio_query` → `POST /synthesis` → WAV, handed back to core for playback.
- Speaker id configurable via `localStorage` (`hermes-ext-voicevox-speaker`); the user's saved rate nudges VOICEVOX's `speedScale`.

## ⚠️ Dependencies

- The core **TTS-engine registration capability** (`window.registerHermesTtsEngine`) — the companion core PR **nesquena/hermes-webui** (TTS-engine registration). No-ops gracefully without it.
- A running **VOICEVOX** server (default `http://127.0.0.1:50021`) — a separate free local app. Fails gracefully if it isn't running.

## Credit
Design reference: closed core PR **nesquena/hermes-webui#4116** (@luperrypf) — the security-clean VOICEVOX integration (hardcoded localhost, no SSRF). This is the extension form, built on the core hook.

## End-to-end verification (live browser, against the core capability)
- ✅ extension loads; **VOICEVOX registers** into the core engine registry
- ✅ **appears in the Settings → TTS Engine dropdown** (`browser / edge / elevenlabs / voicevox`)
- ✅ selecting it + clicking **Listen** routes through the registered `synthesize`, calling **both** VOICEVOX steps (`audio_query` → `synthesis`) — verified via a stubbed loopback fetch (no real server needed for the wiring test); audio played through the core `<audio>` path
- ✅ repo gates green (validate / safety-scan / registry / validator self-test); the loopback `127.0.0.1:50021` literal is accepted by the safety scanner (loopback is allow-listed), so `network_external` is correctly `false`

## Permissions
- `network_external:false` (loopback only); `loopback_sidecar:true`; `storage.owned:["hermes-ext-voicevox-speaker"]`; no DOM, no WebUI API, no cookies, no native host. Hardcoded `127.0.0.1` mirrors the no-SSRF design of the source PR.

### DIFF
--- a/extensions/README.md
+++ b/extensions/README.md
@@ -42,3 +42,5 @@ need it and maintainers agree on the shared contract.
 - `external-app-tab`: pin a compatible self-hosted web app (Grafana, a
   dashboard, or another framable tool) as a tab inside the WebUI via an iframe
   (needs the core `HERMES_WEBUI_CSP_FRAME_EXTRA` knob for external origins).
+- `voicevox-tts`: use a local VOICEVOX server as a TTS engine (registers into
+  Settings → TTS Engine via the core TTS-engine registration hook).
--- a/extensions/voicevox-tts/README.md
+++ b/extensions/voicevox-tts/README.md
@@ -91,17 +91,24 @@ the `extensions/voicevox-tts/` directory.
 This is trusted local code. Current disclosed behavior:
 
 - calls `window.registerHermesTtsEngine(...)` once to add the VOICEVOX engine
-- on synthesis, makes **loopback-only** requests to `http://127.0.0.1:50021`
-  (`/audio_query`, `/synthesis`) — no external network
-  (`network_external: false`; `loopback_sidecar: true`)
-- reads/writes `localStorage` under the single key
-  `hermes-ext-voicevox-speaker`
-- creates **no DOM**
-- does NOT call WebUI HTTP APIs, access cookies, use the filesystem, or use native
-  hosts
+- on synthesis, makes **loopback-only** requests (`/audio_query`, `/synthesis`,
+  `/speakers`) with `credentials:'omit'` — no external network. The server address
+  defaults to `http://127.0.0.1:50021` and is user-overridable, but the override is
+  **strictly validated**: only an http(s) **loopback** host (`localhost`,
+  `127.0.0.0/8`, `::1`) or a safe **root-relative same-origin proxy path** is
+  accepted; any external/protocol-relative URL is rejected and falls back to the
+  loopback default (`network_external: false`; `loopback_sidecar: true`)
+- reads/writes `localStorage` under `hermes-ext-voicevox-speaker` (its own speaker
+  id) and `hermes-ext-voicevox-base` (the server-URL override), and **reads** the
+  shared core key `hermes-tts-voice` (the Settings voice selection)
+- injects **one Settings field** ("VOICEVOX Server URL") next to the core TTS voice
+  selector, and populates the core voice dropdown from `/speakers` — a scoped DOM
+  mutation of the Settings panel only
+- does NOT call WebUI HTTP APIs, access cookies (requests are `credentials:'omit'`),
+  use the filesystem, or use native hosts
 
-The loopback-only address (hardcoded `127.0.0.1`) mirrors the no-SSRF design of
-the source PR.
+The loopback-validation policy preserves the no-SSRF design of the source PR even
+with the configurable URL.
 
 ## Compatibility
 
--- a/extensions/voicevox-tts/README.md
+++ b/extensions/voicevox-tts/README.md
@@ -18,6 +18,18 @@ server.
 - The speaker id is configurable via `localStorage`
   (`hermes-ext-voicevox-speaker`, default 1); the user's saved rate nudges
   VOICEVOX's `speedScale`.
+- **Configurable server URL** — override the VOICEVOX address via a **VOICEVOX
+  Server URL** field injected into Settings → TTS Engine (shown only when VOICEVOX
+  is the selected engine); stored in `localStorage` (`hermes-ext-voicevox-base`,
+  default `http://127.0.0.1:50021`). Absolute URLs and same-origin relative paths
+  (for a reverse proxy) are accepted.
+- **Voice list** — when VOICEVOX is selected, the Voice dropdown is populated from
+  the server's `GET /speakers` (speaker + style names), cached ~30s; changing the
+  server URL refreshes it (debounced).
+
+> Server-URL field, voice-list population, and the configurable base URL were
+> contributed by **@luperrypf** (extensions PR #30), folded into this PR with
+> thanks.
 
 ## Dependency
 
--- a/extensions/voicevox-tts/README.md
+++ b/extensions/voicevox-tts/README.md
@@ -1,0 +1,125 @@
+# VOICEVOX TTS
+
+VOICEVOX TTS is a trusted local Hermes WebUI extension that adds **VOICEVOX** — a
+free, locally-hosted Japanese text-to-speech engine — as a selectable TTS engine.
+Once selected in **Settings → TTS Engine**, both the per-message **Listen** button
+and hands-free **voice mode** synthesize speech through your local VOICEVOX
+server.
+
+## What It Does
+
+- Registers a **VOICEVOX (local)** engine via the core TTS-engine registration
+  capability (`window.registerHermesTtsEngine`), so it appears in the built-in
+  Settings → TTS Engine dropdown alongside Browser / Edge / ElevenLabs.
+- Synthesizes via the standard VOICEVOX two-step flow (`/audio_query` →
+  `/synthesis`) and hands the WAV audio back to core, which plays it through the
+  same `<audio>` lifecycle as the built-in Edge engine (including stop / re-listen
+  in voice mode).
+- The speaker id is configurable via `localStorage`
+  (`hermes-ext-voicevox-speaker`, default 1); the user's saved rate nudges
+  VOICEVOX's `speedScale`.
+
+## Dependency
+
+Requires the core **TTS-engine registration capability**
+(`window.registerHermesTtsEngine`). On an older WebUI without it, the extension
+**no-ops gracefully** (logs a warning; nothing breaks).
+
+You also need a running **VOICEVOX** server (default `http://127.0.0.1:50021`).
+VOICEVOX is a separate, free local application — see the VOICEVOX project. If the
+server isn't running, synthesis fails gracefully (core shows a toast on the Listen
+button / re-listens in voice mode).
+
+## Current Shape
+
+```text
+Hermes WebUI page
+  -> manifest-bundled extension assets
+  -> /extensions/assets/voicevox-tts.js
+  -> window.registerHermesTtsEngine({ id:'voicevox', synthesize })
+  -> Settings -> TTS Engine -> "VOICEVOX (local)"
+  -> synthesize(): POST 127.0.0.1:50021/audio_query -> POST /synthesis -> WAV
+  -> localStorage: hermes-ext-voicevox-speaker (speaker id)
+```
+
+This extension is `static-ui` / manifest-bundle only. It does not add backend
+routes, start a sidecar itself, use native host APIs, or make any **external**
+network calls — it talks **only** to the loopback VOICEVOX server.
+
+## Capabilities
+
+- `manifest-bundle`
+
+## Install For Local Testing
+
+```bash
+# start VOICEVOX (the local app) so 127.0.0.1:50021 is reachable, then:
+cd /path/to/hermes-webui
+HERMES_WEBUI_EXTENSION_DIR=/path/to/hermes-webui-extensions/extensions/voicevox-tts HERMES_WEBUI_EXTENSION_MANIFEST=manifest.json ./start.sh
+```
+
+Open Settings → TTS Engine, choose **VOICEVOX (local)**, then click **Listen** on
+an assistant message (or use voice mode).
+
+## Controls
+
+Also on `window.HermesVoicevoxExtension`:
+
+- `.getSpeaker()` / `.setSpeaker(n)` — VOICEVOX speaker id
+- `.base` — the VOICEVOX server base URL (loopback)
+
+## Disable And Uninstall
+
+Switch the TTS Engine back to Browser/Edge/ElevenLabs, then restart Hermes WebUI
+without `HERMES_WEBUI_EXTENSION_DIR` / `HERMES_WEBUI_EXTENSION_MANIFEST`, or remove
+the `extensions/voicevox-tts/` directory.
+
+## Trust And Permissions
+
+This is trusted local code. Current disclosed behavior:
+
+- calls `window.registerHermesTtsEngine(...)` once to add the VOICEVOX engine
+- on synthesis, makes **loopback-only** requests to `http://127.0.0.1:50021`
+  (`/audio_query`, `/synthesis`) — no external network
+  (`network_external: false`; `loopback_sidecar: true`)
+- reads/writes `localStorage` under the single key
+  `hermes-ext-voicevox-speaker`
+- creates **no DOM**
+- does NOT call WebUI HTTP APIs, access cookies, use the filesystem, or use native
+  hosts
+
+The loopback-only address (hardcoded `127.0.0.1`) mirrors the no-SSRF design of
+the source PR.
+
+## Compatibility
+
+- manifest-bundled extension assets + same-origin serving under `/extensions/`
+- the core TTS-engine registration capability (`window.registerHermesTtsEngine`)
+- a running VOICEVOX server at `127.0.0.1:50021` (loopback is already permitted by
+  the default CSP `connect-src`)
+
+## Verification
+
+```bash
+node scripts/validate-extensions.mjs
+node scripts/scan-extension-safety.mjs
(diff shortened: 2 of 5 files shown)


nesquena-hermes:
@franksong2702 @rodboev — review welcome on this one (and its core dependency, nesquena/hermes-webui#5101, which adds the `registerHermesTtsEngine` hook).

Both **code** and a **hands-on pass** appreciated. Note this one needs a running VOICEVOX server + the core capability PR to exercise live; the registration/dropdown/Listen-routing wiring was verified end-to-end in a browser with a stubbed loopback synth (both VOICEVOX `audio_query` → `synthesis` steps fire). @luperrypf's closed core PR #4116 is the design reference (security-clean, hardcoded localhost / no SSRF).

If anything's off in real use — engine not appearing, playback issues, or behavior the gates wouldn't catch — please flag it. Gates green, rebased on current main.

luperrypf:
Suggested improvements in PR #30: https://github.com/hermes-webui/hermes-webui-extensions/pull/30

- configurable base URL via localStorage
- voice list populated from /speakers
- Settings URL field with MutationObserver

nesquena-hermes:
Thanks @luperrypf — these are great additions, and I've folded all three from your PR #30 into this branch (with a `Co-authored-by` trailer crediting you):

- **Configurable base URL** — `localStorage` key `hermes-ext-voicevox-base` (absolute or same-origin relative path), default `http://127.0.0.1:50021`.
- **Voice list** — populated from `GET /speakers` (speaker + style names), cached ~30s.
- **VOICEVOX Server URL field** — injected into Settings → TTS Engine via a MutationObserver, shown only when VOICEVOX is the selected engine, with a debounced refresh of the voice list on URL change.

I also updated `extension.json` permissions to match the new behavior (`dom.owned` + `mutates_core_views` → true, the new `hermes-ext-voicevox-base` owned key, and `hermes-tts-voice` declared as a shared key) and documented the additions + your credit in the README. Gates green (validate / safety-scan / registry / validator self-test).

I'm closing #30 in favor of this — same content, landed here so it ships as one VOICEVOX entry. Thank you!

nesquena-hermes:

- **Codex (SECURITY):** `baseUrl()` accepted any absolute/protocol-relative URL → could send TTS text to an external host while declaring `network_external:false`. **Fixed** — strict validation: only http(s) **loopback** hosts (`localhost`/`127.0.0.0/8`/`::1`) or a safe root-relative same-origin proxy path; anything else falls back to the loopback default.
- **Codex (SECURITY):** loopback POSTs didn't use `credentials:'omit'`. **Fixed** on all three calls (`audio_query`, `synthesis`, `speakers`).
- **Codex (BREAK):** synthesis ignored the Settings voice selection (`opts.voice`). **Fixed** — `resolveSpeaker(opts)` prefers the validated core voice id.
- **Codex (DISCLOSURE):** README was stale (claimed single key / no DOM / hardcoded URL). **Fixed** to disclose both owned keys, the shared `hermes-tts-voice` read, the Settings-field DOM injection, and the validated URL policy. (Includes @luperrypf's PR #30 improvements, credited via `Co-authored-by`.)

**Verification performed:**
- Functional E2E in a real seeded conversation on an isolated server (not a synthetic harness) — verified the engine registers into core's TTS registry, the Settings URL field injects, and the voice dropdown populates (XSS-safe via `textContent`)

franksong2702:
Hands-on 8787 validation found a real failure path: `VOICEVOX audio_query failed: 500`.

The local VOICEVOX server is running and healthy (`/version` returns `0.25.2`), the extension loads, `VOICEVOX (local)` registers into the TTS engine dropdown, and valid short synthesis works. The failure class is speaker validation:

- VOICEVOX returns 200 for valid speaker ids, e.g. `speaker=1` and `speaker=126`
- VOICEVOX returns 500 for an invalid speaker id, e.g. `speaker=127` or `speaker=9999`
- current `resolveSpeaker(opts)` accepts any non-negative integer from `hermes-tts-voice` / `hermes-ext-voicevox-speaker` without checking that it exists in the live `/speakers` list

So stale numeric browser state can make the UI look configured while every Listen attempt fails with `VOICEVOX audio_query failed: 500`.

Please fix before merge: when VOICEVOX is selected, validate the selected speaker against the fetched `/speakers` ids, clear/fallback stale values, and ideally retry/fallback to a known valid speaker if `/audio_query` returns this invalid-speaker failure. The user-facing error should also include enough context to know the selected speaker id was rejected.

franksong2702:
Additional hands-on finding from @franksong2702: the VOICEVOX voice selection does not persist reliably through the Settings flow.

Repro from the live 8787 bundle:

1. Settings → Preferences
2. Select `TTS Engine = VOICEVOX (local)`
3. The Voice dropdown populates with VOICEVOX speakers/styles
4. Select a concrete voice, e.g. `九州そら (ツンツン)`
5. Click Save Settings
6. Return to the session and click the assistant message speaker icon
7. It fails with `VOICEVOX audio_query failed: 500`
8. Re-open Settings → Preferences: the engine still shows VOICEVOX, but Voice has reverted to `Default system voice` instead of the selected VOICEVOX speaker

Likely integration issue: the extension dynamically repopulates core `#settingsTtsVoice`, but core still treats unknown/non-built-in engines as browser TTS in `_populateTtsVoices()`, and the extension MutationObserver only calls `onEngineChange()` when first hooking `#settingsTtsEngine`. After core/settings save or pane re-entry repopulates the voice select, the extension does not reliably re-own the dropdown, rebind the voice change handler, or restore a valid VOICEVOX speaker id. That leaves stale/empty/invalid `hermes-tts-voice` state, which then reaches `/audio_query` as an invalid `speaker` and produces the 500.

Please fix before merge:

- when engine is `voicevox`, the extension should re-populate/re-own `#settingsTtsVoice` after Settings save, section switches, and any core `_populateTtsVoices()` repaint
- bind `#settingsTtsVoice` changes from the VOICEVOX list to persist a valid speaker id (`hermes-tts-voice`, and/or the extension speaker key consistently)
- validate persisted speaker ids against the live `/speakers` ids and fallback/clear stale values
- avoid showing `Default system voice` while `TTS Engine = VOICEVOX (local)`; it should show `Default speaker` or a concrete VOICEVOX speaker

This currently reproduces as a user-visible settings persistence failure plus the `audio_query failed: 500` playback failure, so I would not merge #29 yet.

nesquena-hermes:
## Both findings fixed + verified live 🙏 @franksong2702 @rodboev

Thanks Frank — both of your hands-on findings reproduced and are fixed on the new head (`154c82c`), rebased clean onto `main`.

### 1. `audio_query failed: 500` on a stale/invalid speaker id
Root cause confirmed: `resolveSpeaker()` accepted any non-negative integer, so a numeric id no longer present on the live engine build reached `/audio_query` as an invalid speaker → 500, dead Listen button.

Fix:
- New `pickSpeaker(opts)` validates the candidate against the fetched `/speakers` ids before synthesis. If the persisted id isn't a real speaker, it falls back to the first real speaker **and clears the stale `hermes-tts-voice` / `hermes-ext-voicevox-speaker`** so the UI self-heals.
- `synthesize()` retries once with a known-valid speaker if `/audio_query` returns the invalid-speaker 500.
- Error text now includes the rejected speaker id (`VOICEVOX audio_query failed: 500 (speaker 127)`).

**Verified live** against a mock VOICEVOX server that returns 200 for valid ids and 500 for invalid (your exact failure class): planted `hermes-tts-voice=127` (not in `/speakers`), called the production playback path → it now returns real audio (44-byte WAV) and clears the stale value. Previously this hard-failed.

### 2. Voice selection not persisting through the Settings flow
Root cause confirmed: core's `_populateTtsVoices()` treats any non-built-in engine as browser TTS, so on Save / section-switch / panel re-open it wiped the VOICEVOX list back to "Default system voice" and dropped the selection.

Fix:
- The extension now **wraps `window._populateTtsVoices`** so VOICEVOX owns the voice dropdown whenever it's the active engine, delegating to core otherwise. The wrapper is re-installed every observer tick (core reassigns the fn on each `loadSettingsPanel`).
- The voice `change` is bound to persist a valid speaker id; stale ids that aren't valid VOICEVOX speakers are dropped on repopulate.
- The "VOICEVOX Server URL" field visibility is now synced to the effective engine every tick (it previously stayed hidden because the one-shot `onEngineChange()` ran before core set the engine value).

**Verified live**: selected `九州そら (ツンツン)` (id 126) → ran a full `loadSettingsPanel()` re-init (the most aggressive repaint) → selection survives, dropdown stays VOICEVOX-owned ("Default speaker" first option, not "Default system voice"), and the real playback path synthesizes successfully from the persisted voice. Screenshot confirmed.

### Codex security/breakage gate
Ran the extension Codex gate. First pass flagged one disclosure gap — the Settings re-ownership reads `localStorage['hermes-tts-engine']` directly, which wasn't in `permissions.storage.shared_webui_keys`. Added that key + disclosed it (and the stale-voice self-heal) in the README. **Re-gate: SAFE TO SHIP** (no security issue, no breakage, disclosure honest; loopback URL validation, `credentials:'omit'`, and `textContent` voice labels all confirmed).

Rebased onto `main` (0 behind), validators + safety scan green. Tagging you both for one last look before merge.

franksong2702:
Residual hands-on blocker from the 8787 bundle: `VOICEVOX audio_query failed: 500 (speaker 2)` still reproduces on real assistant-message playback.

I isolated this from the earlier invalid-speaker issue. `speaker=2` itself is valid on the local engine:

- local VOICEVOX `/version` returns `0.25.2`
- `/speakers` includes `四国めたん / ノーマル` with style id `2`
- short `/audio_query` requests with `speaker=2` return 200

The remaining failure class appears to be long / Markdown-heavy assistant text being sent to VOICEVOX in one `audio_query` request:

- short Japanese text with `speaker=2` -> 200
- short English text with `speaker=2` -> 200
- short Chinese text with `speaker=2` -> 200
- around 4.5k Chinese chars -> 500
- around 6.4k Markdown/list/URL/code-like text -> 500

So this is not “speaker 2 is invalid”. The extension currently passes the whole assistant message through `audio_query?text=...` at once, and long final answers can still fail even with a valid speaker. That is exactly the user-visible path for Hermes answers: click the assistant message speaker icon on a long response, and playback fails.

Please fix before merge by handling long assistant messages explicitly. A reasonable path would be:

- normalize the input before synthesis (flatten or remove Markdown noise, code blocks, URLs, etc.)
- split text into bounded chunks by sentence/length before calling `/audio_query`
- synthesize/play chunks sequentially, or concatenate audio if that is the chosen implementation
- make the error distinguish “VOICEVOX rejected this text/length” from “selected speaker is invalid”

If the first version intentionally only supports short text, the UI should cap the text and disclose that behavior clearly. But as-is, a normal long Hermes final answer still hits `audio_query` 500, so I would not merge #29 yet.

nesquena-hermes:
## Long-answer playback fixed + a ReDoS the fix surfaced 🙏 @franksong2702 @rodboev

Thanks Frank — your isolation was spot-on. This was **not** an invalid-speaker problem; VOICEVOX's `/audio_query` 500s on long / Markdown-heavy text (~4k+ chars) even with a valid speaker, which is exactly the "click Listen on a long final answer" path. Fixed on the new head (`0398057`), rebased clean onto `main`.

### The fix
- **`normalizeForSpeech()`** — strips what shouldn't be read aloud anyway: fenced + inline code, image/link URLs (keeps link labels), heading/list/emphasis/blockquote markers, collapses whitespace.
- **`chunkText()`** — splits into ≤1800-char pieces on sentence/newline boundaries (CJK `。！？` + ASCII `.!?`), hard-splitting a pathological single sentence so no chunk ever exceeds the cap.
- **`synthesize()`** — normalize → chunk → synthesize each chunk sequentially (locking in a speaker corrected on the first slice) → **`concatWavs()`** stitches the returned WAV `data` payloads into one clip core plays.
- Errors now carry speaker id **and** char count, so a 500 reads as invalid-speaker vs too-long; the invalid-speaker retry only fires on short slices.

### Verified end-to-end (mock VOICEVOX reproducing your exact behavior: valid speaker 2, 500 on >2000-char text)
Drove the real production path (`window._hermesTtsSynth`) on a seeded **7616-char** Markdown answer:
- **Control (proves the bug):** the whole text in one `/audio_query?speaker=2` → **500**.
- **Extension path:** **succeeds** — 5 chunks of `[1630, 1751, 1798, 1734, 638]` chars (all <2000), concatenated into one valid WAV.
- Invalid-speaker on short text still self-heals; normalization confirmed to strip `` `code` ``, URLs, and `**bold**` markers.

### Codex security gate — caught a real ReDoS, now fixed
The gate flagged my first Markdown-strip implementation as ReDoS-vulnerable (superlinear backtracking on crafted `[x](` openers — assistant/session text is attacker-influenceable in the authed tab). Rewrote it as **`stripMarkdownLinks()`**, a single left-to-right scan that probes only a bounded 4096-char slice per bracket — genuinely O(n) (1M malformed chars: ~7s → ~135ms). **Re-gate verdict: SAFE TO SHIP** (Codex re-verified linear timing + no other issues).

This one had real fixes across two areas, so I'm **not** self-merging — tagging you both for another hands-on pass. Frank, if you can re-run the long-answer Listen case on the 8787 bundle that'd be the decisive check.

nesquena-hermes:
Rebased onto current `main` (0 behind, MERGEABLE); validators + safety scan green. No code change since the long-answer chunking + ReDoS-safe normalizer fix — just keeping the branch current. Still awaiting your hands-on Listen re-test on the long-answer path when you get a chance 🙏 @franksong2702

franksong2702:
Thanks for the continued fixes here. I re-tested the current PR branch on the live 8787 bundle with a local VOICEVOX server.

Current result:

- The basic integration path is working: the extension loads, `VOICEVOX (local)` appears in Preferences, the Settings fields are visible, and short assistant messages can play successfully.
- The earlier invalid/stale speaker-id class looks addressed in the paths I tested.
- The remaining blocker is long assistant-answer playback. Long Japanese assistant responses still do not start playing in a reasonable amount of time. In testing, short replies can be heard, but a longer Japanese response appears to spend a long time generating before any audio is heard.
- This creates a confusing user experience: the speaker button can look active/pending, but there is no audible output for a long time. In earlier runs, audio could start much later, which felt like delayed queued playback rather than an immediate Listen action.

My read is that this is no longer just the original `/audio_query` 500 bug. The extension now normalizes/chunks before calling VOICEVOX, but the current implementation still returns one complete concatenated WAV to core. Since core only starts playback after the extension returns the full audio buffer, long final answers can still feel broken even if the individual VOICEVOX requests eventually succeed.

So I do not think #29 is ready to merge yet.

For this to be usable on normal Hermes assistant answers, long text needs an explicit playback story: either the extension/core path needs progressive chunk playback with reliable cancellation, or another approach that lets users hear the first chunk promptly and stop pending playback predictably. Only supporting short messages, silently taking a long time, or capping to the first N characters would not be enough for this use case.

This extension is more complex than it first looked, so I think we should keep this PR open and treat long-answer playback as the remaining acceptance criterion before merge.

nesquena-hermes:
Thanks for the thorough hands-on re-test @franksong2702 — and your read is correct. I verified it against the core registered-TTS-engine contract:

- Core invokes an extension engine via `window._hermesTtsSynth(engine, text, opts)`, which must resolve to **one complete audio buffer**, then hands it straight to `_playAudioBuf(buf)` (`static/ui.js`). Playback only begins after the full buffer resolves.
- So for a long Japanese answer, the extension's per-chunk VOICEVOX requests can each succeed, but core still waits for the whole concatenated WAV before any audio is heard — exactly the "speaker looks active/pending, no audible output for a long time" UX you saw. This is no longer the original `/audio_query` 500; it's the single-buffer contract.

That means progressive first-chunk-plays-immediately playback with reliable cancellation **cannot be done from the extension alone today** — it needs a core capability: a streaming/progressive synth contract where the engine can yield audio chunks as they're ready and core plays + lets the user stop mid-stream. Capping to the first N characters or silently taking a long time isn't acceptable, agreed.

So I'll keep #29 **open** and treat long-answer progressive playback as the remaining blocker, and I'm logging "progressive/streaming TTS-engine synth contract" as a capability gap on the extension-capability umbrella (#40) so the core side is tracked. Once that capability lands, this extension can adopt it and the long-answer path becomes usable. Appreciate you catching that the completion-buffer path is the real wall here.

nesquena-hermes:
### Maintainer status refresh — 2026-07-30

Rechecked this against current core source rather than assuming the old blocker had changed.

The progressive-output contract is still missing: registered extension engines resolve one complete audio buffer through `_hermesTtsSynth()` before `_playAudioBuf()` is called in both the per-message and auto-read paths. There is still no incremental first-audio sink, cancellation/epoch contract, or progressive queue interface that a VoiceVox engine can implement honestly.

So this PR remains **blocked on core capability, not on contributor action**. I am not asking for a rebase or more patching of this branch yet; doing that now would only churn a conflicting PR without making progressive playback possible. The current capability status is also recorded in #40. Once core exposes a stable incremental TTS contract, this can be refreshed onto current `main` and gated against that API.

PR #31, opened 2026-06-29
Refine mobile conversations drawer close affordance

santastabber:
## Summary
- keep the mobile Conversations extension button visible above the open drawer as a close X
- preserve the same lower-right floating position used by the normal button
- update extension metadata/docs to describe the close affordance and DOM reparenting behavior

## Verification
- `node --check extensions/mobile-conversations/assets/mobile-conversations.js`
- `python3 -m json.tool extensions/mobile-conversations/extension.json >/dev/null`
- `python3 -m json.tool extensions/mobile-conversations/manifest.json >/dev/null`
- `git diff --check`

### DIFF
--- a/extensions/README.md
+++ b/extensions/README.md
@@ -19,5 +19,5 @@ need it and maintainers agree on the shared contract.
 
 - `desktop-companion`: trusted local Desktop Companion entry and first
   sidecar-class extension candidate.
-- `mobile-conversations`: phone-only floating Conversations button and
-  long-press shortcuts for the existing Hermes WebUI mobile drawer.
+- `mobile-conversations`: phone-only floating Conversations button, same-location
+  drawer close X, and long-press shortcuts for the existing Hermes WebUI mobile drawer.
--- a/extensions/mobile-conversations/README.md
+++ b/extensions/mobile-conversations/README.md
@@ -2,20 +2,21 @@
 
 Mobile Conversations Button is a trusted local Hermes WebUI extension that adds
 a phone-only floating **Conversations** button near the chat composer. It opens
-the existing mobile conversations drawer and provides a long-press shortcuts
-menu for common conversation navigation actions.
+the existing mobile conversations drawer, then stays in the same floating
+position as an **X** above the open drawer so it can be closed again. It also
+provides a long-press shortcuts menu for common conversation navigation actions.
 
 ## What It Does
 
 - Adds a phone-width-only floating Conversations button inside the chat messages
   shell.
-- Short tap opens or closes the existing mobile conversations/sidebar drawer.
+- Short tap opens the existing mobile conversations/sidebar drawer.
 - Long press opens a menu with shortcuts for:
   - New conversation
   - Open sidebar
   - Go to top
   - Go to last message
-- Hides the floating button while the mobile drawer is already open.
+- Keeps the same button in its normal floating position above the open mobile drawer as an **X** close affordance.
 - Leaves desktop-width layouts unchanged.
 
 ## Current Shape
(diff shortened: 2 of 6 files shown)


nesquena-hermes:
## Deep review — verified end-to-end on a real mobile viewport ✅

Reviewed the **final head** (`d82da72`) by serving the PR branch's extension against a clean `origin/master` core worktree on an isolated port (8789), driving it in a headless Chrome at **412×915 (phone)** through a 16-message conversation, and vision-checking the pixels. Recommendation: **good to merge** — one optional design note below, no blockers.

### What the PR changes
Previously the floating Conversations button was `display:none` while the drawer was open. Now it stays visible, reparents to `document.body`, morphs into an **X**, and sits at the same lower-right floating position as a close affordance. Clean, well-scoped refinement to the author's own extension.

### Functional verification (all passed)
- **Open → X → close round-trip.** Tap opens the drawer; button reparents to `body`, `position:fixed`, `z-index:240`, icon swaps to two `<line>`s (X), `aria-label`→"Close conversations", `aria-haspopup` removed. Tapping the X closes the drawer and the button returns to `.messages-shell` (`z-index:12`, message glyph, `aria-haspopup:menu` restored).
- **Stacking is correct in the real DOM, not just on paper.** Mobile `.sidebar.mobile-open` is `z-index:200`; the reparented X is `z-index:240`. `elementFromPoint()` at the button center resolves to the button (its SVG line), so the X is genuinely topmost and clickable above the full-screen drawer. Screenshot confirms it renders fully visible, unclipped, above the drawer.
- **Core-initiated close recovers.** Opened via the button, then closed via `window.closeMobileSidebar()` (simulating a hamburger/session tap) — the `MutationObserver` on `.sidebar[class]` fires `syncButton()` and the button correctly returns to `.messages-shell`. No orphaned body-level button.
- **Long-press / context-menu correctly suppressed while open.** The new `pointerdown`/`contextmenu` guards (`if (sidebarOpen()) return`) prevent the shortcuts menu from opening when the button is acting as a close X. Verified the menu does NOT open on a touch long-press while the drawer is open.
- **Menu still works when closed.** Long-press while closed still opens the 4-item shortcuts menu (New conversation / Open sidebar / Go to top / Go to last message).
- **Re-render survival.** With the drawer open (button parked in `body`), wiped `#messages` innerHTML to simulate `renderMessages()` — button survives, and closes back into `.messages-shell`.
- **No closed-state collision.** FAB does not overlap `#scrollToBottomBtn` or `#jumpToSessionStartBtn`.
- **Desktop unchanged.** At 1440px the button is `display:none`. No console errors across the whole flow.

### Gates + trust audit (all clean)
- `node --check`, `validate-extensions.mjs`, `scan-extension-safety.mjs`, `test-extension-validator.mjs`, `git diff --check` — all pass.
- **Permissions block is honest.** Static audit of `mobile-conversations.js`: 0 `fetch`, 0 `localStorage`, 0 `/api/`, 0 `eval`/`Function`, 0 external URLs (only the `w3.org/2000/svg` namespace literal). So `webui_api:{read:[],write:[]}`, `network_external:false`, `storage:{owned:[],shared:[]}` remain accurate. `dom.mutates_core_views:true` is now genuinely warranted since the button is appended to `document.body` while open — good that the disclosure tracks the new reparenting behavior, and the README's "temporarily reparents it to `document.body`" line documents it transparently.

### One optional design note (non-blocking)
When the drawer is open there are now **two** close affordances on screen at once: core's header **✕** (top-right of the drawer) and this floating **✕** (bottom-right). Both close the drawer, so it's not a bug — but it's a small bit of redundancy a user might notice. Two ways to read it: (a) the bottom-right X is reachable with the same thumb that opened it (nice for one-handed use, which is the whole point of the FAB), so the redundancy is arguably justified; (b) if you'd rather avoid two X's, the alternative is keeping the original hide-on-open behavior. I lean toward keeping it as-is for the thumb-reach ergonomics — just flagging it so the choice is deliberate.

Nice, tight change. The guard logic and observer-driven reparenting are handled correctly, and the permission disclosure stays honest. 👍

— verified in a long conversation on an isolated server at phone width; drove open/close/core-close/long-press/menu/re-render and vision-checked the drawer-open state.

nesquena-hermes:
Merged in `2e98a879` (squash). Thanks @santastabber — clean refinement to your own extension. 🙏

Verified end-to-end before merge on an isolated server at phone width (412px): open → X → close round-trip, correct stacking above the drawer (`z:240` over the `z:200` drawer, `elementFromPoint` confirms the X is topmost/clickable), core-initiated-close recovery via the sidebar MutationObserver, long-press suppressed while open, shortcuts menu intact when closed, re-render survival, desktop hidden, no console errors. All repo gates green and the `permissions` block re-audited honest (the new `document.body` reparent is correctly disclosed via `dom.mutates_core_views`).

The registry will pick up `mobile-conversations` v0.1.1 on the next push-to-main deploy.

Issue #34, opened 2026-07-01
🔧 Capability retrofit + docs pass across the extension library


nesquena-hermes:
# 🔧 Capability retrofit + docs pass across the extension library

The library has 19 merged entries, and each worthwhile retrofit ships as its own one-extension PR through the full pipeline (validators + `scan-extension-safety` + Codex gate + browser QA). This is value-driven work, not a requirement to migrate every entry mechanically.

## Current state — 2026-08-27

- Installed-manifest settings parity is complete for `mobile-haptics` (#81) and `external-app-tab` (#83), with real-Core installed-artifact coverage.
- The scoped Configure contract shipped in Core #7116, its real-Core certification merged in #86, and Typography became the first real consumer in #88.
- Typography now opens its existing editor through Core's Installed-page Configure affordance instead of a permanent launcher. Its imported-font storage, preview behavior, and programmatic API remain extension-owned.
- Theme Creator became the second Configure consumer in #89; its separate empty-background visibility fix shipped in #92. No third Configure consumer is selected by default.
- A3 message actions remain in Core PR [#7245](https://github.com/nesquena/hermes-webui/pull/7245). After independent Core review and merge, #76 must certify the merged Core revision before a separate Message Pins migration.
- Collection-like state (pins, favorites, themes, dictionaries, uploaded assets) stays extension-owned. A next consumer should be selected only when a sanctioned capability removes a concrete compatibility or UX problem.

## Capabilities to adopt

1. **Sanctioned settings system** (`settings_schema` + `permissions.storage.owned` + `window.HermesExtensionSettings`) — replace hand-rolled localStorage settings panels so options render natively in Settings → Extensions → [ext]. (See #33.)
2. **Skin base-scheme** `scheme:'light'|'dark'` — already adopted by `e-ink-skin` + `skin-pack`; any future skin entry should use it, never per-token CSS workarounds.
3. **Best-practice hardening already proven in-review** — absolute same-origin `fetch` (never route-relative `window.api`), ReDoS-safe text parsing, MutationObserver scoping/disconnect, off-origin-media sanitization on any HTML clone, honest permission disclosure.

## ✅ Enabling blocker resolved — safety-scan vs settings_schema storage.owned contract

The original direct contract conflict was:

- **Core** honors `settings_schema` only when `permissions.storage.owned === true` (boolean) — `api/extensions.py:_manifest_entry_storage_owned` returns `storage.get("owned") is True`.
[...]

Issue #33, opened 2026-07-01
Feature: add settings_schema support to gallery extensions for in-Settings customization


ChonSong:
Gallery extensions like `assistant-avatar`, `custom-avatar`, `theme-creator` etc. could benefit from the extension settings system (`permissions.storage.owned` + `settings_schema` in manifest.json).

Currently, extensions that want user-configurable options (colors, visibility toggles, behavior flags) have to implement their own localStorage-based settings panels from scratch. The WebUI already supports a sanctioned settings system via `window.HermesExtensionSettings`, `settings_schema`, and `permissions.storage.owned` — see docs/EXTENSIONS.md.

**Request:** Add `settings_schema` + `permissions.storage.owned` to gallery extensions that would benefit from settings, starting with expression-based avatars (assistant-avatar, custom-avatar). This would surface color pickers, toggle switches, etc. directly in Settings → Extensions → [extension] → settings, consistent with how built-in extension settings work.

**Reference:** The `assistant-avatar` (2D Canvas face floating in the corner) already has a working local implementation of localStorage-backed config + a JS settings popover. Happy to contribute a PR porting that to the sanctioned `settings_schema` pattern.

**Relevant docs:** https://github.com/hermes-webui/hermes-webui/blob/main/docs/EXTENSIONS.md (lines ~168-220 cover settings_schema)

nesquena-hermes:
Thanks @ChonSong — this is a good call and it's in scope. I confirmed the sanctioned settings system is live in core: `static/extension_settings.js` exposes `window.HermesExtensionSettings` (with `settingsForExtension()` / `storageForExtension()`), it's wired into `Settings → Extensions → [extension]` in `panels.js`, and `docs/EXTENSIONS.md` (§ settings_schema, ~L179–215) documents the manifest shape. So porting gallery extensions off ad-hoc localStorage panels onto `settings_schema` is exactly the right direction.

A few specifics so a PR lands clean the first time:

- **`settings_schema` is honored only when `permissions.storage.owned === true`** (exactly `true`, not just a key list) — that's the gate in core. Declare it alongside the schema.
[...]

PR #36, opened 2026-07-01
mobile-haptics: adopt sanctioned settings_schema for the enable toggle (retrofit #34)

nesquena-hermes:
## Exemplar capability retrofit — mobile-haptics → sanctioned `settings_schema` (issue #34)

First retrofit in the #34 effort, and the pattern-setter for the rest. Unblocked by #35 (gates now accept `storage.owned: true`).

### What changed
- **`extension.json`**: `permissions.storage.owned` → `true` (was `["hermes-ext-haptics-enabled"]`) + a `settings_schema` with one boolean field `enabled` (default `true`). `owned: true` is the exact form core requires to honor `settings_schema`.
- **`mobile-haptics.js`**: `enabled()` / `setEnabled()` now read/write through `window.HermesExtensionSettings.settingsForExtension("mobile-haptics")` when core supports it (`.supported`), falling back to the legacy `hermes-ext-haptics-enabled` localStorage key on older core. Default ON on both paths. The turn-complete state machine (busy→idle, `queue` holding-state guard, `MIN_BUSY_MS` floor) is unchanged.
- **README**: documents the native Settings → Extensions toggle + the legacy fallback.

### User-visible win
The on/off preference now renders as a **native toggle in Settings → Extensions → Mobile Haptics** ("Vibrate when a turn finishes") instead of an invisible localStorage key — consistent with how built-in extension settings work.

### Verified (both read paths drive behavior, in a browser against master)
- **Settings store** (`supported:true`): `enabled:true` → buzz; `enabled:false` → no buzz; store-`false` **wins over** a stale legacy `"1"` (store is authoritative, not double-consulted).
- **Legacy fallback** (older core, no settings system): default → buzz; key `"0"` → no buzz.
- `validate-extensions` + `scan-extension-safety` + validator self-test all green; Codex gate running.

This is the template for the remaining retrofits in #34 (custom-avatar, theme-creator, model-favorites, …), one-per-PR. Since it changes a shipped extension's behavior surface, **not self-merging** — tagging @franksong2702 @rodboev for review. Refs #34, #33.

### DIFF
--- a/extensions/mobile-haptics/README.md
+++ b/extensions/mobile-haptics/README.md
@@ -8,7 +8,7 @@ and set the device down, you get a physical "it's done" cue.
 
 - Watches for the end of an assistant turn and triggers a short
   `navigator.vibrate()` buzz.
-- Opt-in preference stored locally (`hermes-ext-haptics-enabled`); on by default
+- Opt-in preference, on by default; managed via the native Settings → Extensions toggle (sanctioned extension-settings store, legacy `hermes-ext-haptics-enabled` localStorage fallback on older core)
   where vibration is supported.
 - Ignores sub-100ms flickers; the real "a turn happened" gate is observing a
   genuine busy action (stop/steer/interrupt), so only real turns buzz.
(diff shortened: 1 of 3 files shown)


nesquena-hermes:
### Codex gate: SAFE TO SHIP ✅

Ran the extension Codex security/breakage gate on the retrofit head. **No findings.** It confirmed:
- `enabled()` uses `HermesExtensionSettings` as the **authoritative path when supported and does not consult the legacy key in that path**; `setEnabled()` writes through settings when available with the older-core fallback.
- Permissions honest: `storage.owned: true` + `device_vibration: true`, no WebUI API writes, no external network, no navigation, no DOM-mutation claims.
- No HTML injection / eval / cookie access / dangerous patterns.
- `validate-extensions` + `scan-extension-safety` + registry generation + `node --check` all pass.

Tagging @franksong2702 @rodboev for the final review — this is the exemplar for the #34 retrofit series, so worth a look at the pattern before I apply it to the other extensions.

PR #42, opened 2026-07-04
Add chat-tiling extension (migrated from core #5312, @ChonSong)

nesquena-hermes:
## Add chat-tiling extension (migrated from core PR #5312)

Adds a **chat-tiling** client-side extension: work with multiple conversations at once in a resizable tile grid (2-column split / 4-corner 2×2 / 3×2), each tile with its own composer, model selection, message list, and live streaming; maximize/restore individual tiles; configurable default layout.

### Why here (not core)
This was originally proposed as [nesquena/hermes-webui#5312](https://github.com/nesquena/hermes-webui/pull/5312) by **@ChonSong**. It's entirely self-contained and client-side (manifest + `assets/tiling.js`, no core changes, no server component, no network access), so the extensions repository is its proper home — the core repo intentionally bundles no extensions. We're migrating it here rather than merging into core, preserving full authorship credit.

### Contents
- `extensions/chat-tiling/manifest.json` — extension manifest (id, permissions, `settings_schema` for default-layout + auto-tile). Permissions are tight: DOM-owned, storage-owned, `webui_api` read-only on `session`/`sessions`, no network/filesystem/sidecar/native-host.
- `extensions/chat-tiling/assets/tiling.js` — the extension (261 lines).
- `extensions/chat-tiling/README.md` — what it does + install + settings + credit.
- Listed in the `extensions/README.md` index.

### Credit
Authored by @ChonSong (core PR #5312). Migrated with authorship preserved (`Co-authored-by` on the commit).

Co-authored-by: ChonSong <ChonSong@users.noreply.github.com>

### DIFF
--- a/extensions/README.md
+++ b/extensions/README.md
@@ -44,3 +44,6 @@ need it and maintainers agree on the shared contract.
 - `external-app-tab`: pin a compatible self-hosted web app (Grafana, a
   dashboard, or another framable tool) as a tab inside the WebUI via an iframe
   (needs the core `HERMES_WEBUI_CSP_FRAME_EXTRA` knob for external origins).
+- `chat-tiling`: work with multiple conversations at once in a resizable tile
+  grid (2-column / 4-corner 2×2 / 3×2), each tile with its own composer, model,
+  messages, and live streaming; maximize/restore individual tiles.
--- a/extensions/chat-tiling/README.md
+++ b/extensions/chat-tiling/README.md
@@ -1,0 +1,34 @@
+# Chat Tiling
+
+Chat Tiling is a client-side Hermes WebUI extension that lets you work with
+multiple conversations at once in a resizable tile grid, instead of one session
+at a time.
+
+## What It Does
+
+- Adds tiling layouts to the chat area: **2-column** split, **4 corners** (2×2
+  grid), and **6 tiles** (3×2 grid).
+- Click a session in the sidebar to open it in the focused tile.
+- Each tile has its own independent composer, model selection, message list, and
+  live streaming — so you can run and watch several conversations side by side.
+- Maximize / restore an individual tile.
+- Remembers your preferred default layout (configurable in the extension's
+  settings).
+
+## Install
+
+Copy the `chat-tiling` folder into your Hermes WebUI extensions directory
+(`HERMES_WEBUI_EXTENSION_DIR`), then enable it from Settings → Extensions. It is
+entirely client-side — no core changes, no server component, no network access.
+
+## Settings
+
+- **Default layout** — which grid to open with (2-column / 4 corners / 6 tiles).
+- **Auto-tile** — whether to enter a tiled layout automatically.
+
+## Credits
+
+Originally authored by [@ChonSong](https://github.com/ChonSong) and contributed
+to Hermes WebUI as PR nesquena/hermes-webui#5312. Migrated here to the extensions
+repository because it is a self-contained client-side extension (the right home
+for it) rather than a core change.
--- a/extensions/chat-tiling/README.md
+++ b/extensions/chat-tiling/README.md
@@ -1,19 +1,22 @@
 # Chat Tiling
 
-Chat Tiling is a client-side Hermes WebUI extension that lets you work with
-multiple conversations at once in a resizable tile grid, instead of one session
-at a time.
+Chat Tiling is a client-side Hermes WebUI extension that lets you view several
+conversations at once in a resizable tile grid, instead of one session at a time.
 
 ## What It Does
 
 - Adds tiling layouts to the chat area: **2-column** split, **4 corners** (2×2
-  grid), and **6 tiles** (3×2 grid).
+  grid), and **6 tiles** (3×2 grid). Switch layouts from the toolbar buttons in
+  the app title bar, or with **Ctrl/⌘ + Alt + 1 / 2 / 4 / 6**.
 - Click a session in the sidebar to open it in the focused tile.
-- Each tile has its own independent composer, model selection, message list, and
-  live streaming — so you can run and watch several conversations side by side.
+- Each tile shows that conversation's transcript (rendered with the same
+  formatting as the main chat — bold, code, code blocks, links). The **focused**
+  tile is wired to the shared composer, model picker, and live stream, so you send
+  and watch in whichever tile you've clicked into; click another tile to switch
+  focus. A small dot in a tile header marks a conversation that's actively running.
 - Maximize / restore an individual tile.
-- Remembers your preferred default layout (configurable in the extension's
-  settings).
+- Remembers your last-used layout locally and can restore it (or your configured
+  default) on load.
 
 ## Install
 
@@ -23,8 +26,13 @@ entirely client-side — no core changes, no server component, no network access
 
 ## Settings
 
-- **Default layout** — which grid to open with (2-column / 4 corners / 6 tiles).
-- **Auto-tile** — whether to enter a tiled layout automatically.
+- **Default layout** — which grid the auto-tile-on-load restore opens when there's
+  no saved last-used layout (2-column / 4 corners / 6 tiles).
+- **Enter tiling automatically on load** — when on, the grid opens on page load
+  (your last-used layout, or the default above); when off, the app loads untiled
+  and you enter tiling from the toolbar.
+- **Show tile count badges in sidebar** — show a small badge on a sidebar session
+  row indicating how many tiles currently hold that conversation.
 
 ## Credits
 
--- a/extensions/chat-tiling/assets/tiling.js
+++ b/extensions/chat-tiling/assets/tiling.js
@@ -124,6 +124,25 @@ function _extRenderMd(raw){
   return null;
 }
 function _extEscape(s){const d=document.createElement('div');d.textContent=String(s==null?'':s);return d.innerHTML}
+// Neutralize off-origin resource loads in rendered markdown. Core renderMd()
+// permits https images (<img src=...>), which would fire off-origin requests on
+// EVERY tile render even though this extension declares network_external:false.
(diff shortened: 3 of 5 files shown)


nesquena-hermes:
## Full QA gate review — parking this open (not merging)

Ran the complete extension QA playbook on this: static gates, an empirical live-drive on an isolated server, and the dual **Codex-ext** (security/breakage) + **Fable** (UX/architecture) senior gate. Both gates came back against merge — Codex `KICK BACK TO AUTHOR`, Fable `UX-REWORK` — and after digging in I agree. The concept is good and near-term useful, but it can't be made solid **as an extension** against the current core. Details below so this is actionable if we ever revisit it.

### What I fixed on the branch (verified)

These were real defects and are now addressed (commit `ef696b6`):

1. **Primary interaction was dead on arrival.** The sidebar-click flow targeted `#sessionSidebar` / `[data-session-id]`, but current core uses `#sessionList` / `.session-item[data-sid]` — so the click listener *never attached* and clicking a session never filled a tile (confirmed live: `extWiredFlag: "container-not-found"`). This came from fork-DOM selectors that drifted from core. Rewired to the real container/attribute + the observer now disconnects instead of watching `document.body` forever.
2. **Messages rendered as raw markdown plaintext** — `renderMsgs()` relied on `window._createMessageElement`, which doesn't exist in core, so every tile fell through to a `textContent` fallback. Rewired to core's real `renderMd()`; verified live that bold / inline code / code blocks / links now render.
3. **Missing `extension.json`** — the entry shipped only a `manifest.json`, so it was invisible to the whole toolchain (validate-extensions / scan-extension-safety / registry generator all discover via `extension.json`). It was the only entry lacking it. Added it as the source of truth; slimmed `manifest.json` to the runtime shape. Now: `ok chat-tiling`, 14 entries validate + safety-scan + generate cleanly.
4. **Two of three settings were dead** (`default_layout`, `auto_tile` were never read). Wired both.
5. **Security:** `renderMd()` emits ``, so tile rendering fired off-origin requests while `network_external:false`. Now strips off-origin `src`/`srcset`/`poster`.
6. **`badge()` CSS-selector injection** (raw session id → `SyntaxError`). Now `CSS.escape`d + correct append target.
7. **Destructive/state-corrupting paths:** the always-visible Close button wiped the real session (`S.*`) when clicked before tiling; closing a tile silently *killed* in-flight agent runs. Now guarded, and tile-close **detaches** streams instead of cancelling them.
8. Corrected stale docs (keyboard shortcuts, the "independent composer per tile" claim) + rewrote the README to describe the real focus-switching model.

### The limitation that keeps it parked

The root issue is architectural, not a bug list. Core WebUI is built as a **singleton** — one global state `S`, one transcript container `#msgInner`, one composer, one live stream, and `loadSession()` as the only safe switch path. Chat-tiling wants **N** conversations at once, and with no plural primitive to build on it **impersonates the singleton**: it physically moves the `#msgInner` id between tiles and reassigns `S` on every focus change, with a 500ms poll copying state back.

That model can't be made robust from extension-space:
- **Non-focused tiles are frozen snapshots** — only the focused tile is live, so "watch several conversations side by side" isn't actually delivered.
- **Mid-stream focus changes can misroute transcript content**, because core streams into whichever element currently owns the `#msgInner` id / the `S.messages` array.

These aren't patchable without core changes — the extension is forging an identity that core assumes is unique.

### What a supported version would need (for the record, not a commitment)

A robust chat-tiling would need core to expose plural-session primitives, ideally as sanctioned registration hooks (same shape as `registerHermesSkin()` / `registerHermesTtsEngine()`):

- **A** — a session-open interception hook (a cancelable event / `registerHermesSessionOpenHandler`), so an extension routes opens without racing core navigation. *(The original core PR #5312 actually had this — a hook inside `loadSession()`.)*
- **B** — `renderTranscript(container, messages, opts)`: render a transcript into an arbitrary element via the real markdown/tool-card pipeline, without touching `S` / `#msgInner`.
- **C** — `subscribeSessionStream(sid, handlers)`: a per-session live subscription decoupled from the single "active" stream (this is what makes non-focused tiles go live).
- **D/E** — per-session `sendMessage(sid, …)` + session-scoped state reads (fully independent tiles = core genuinely becoming multi-session, which is a product decision in its own right).

A + B alone would make a legitimate, honest extension (focused tile live, others read-only views — the VS-Code-split model). Everything past that is core becoming multi-session.

### Disposition

**Leaving this open.** There's no reasonable path to shipping it as a solid extension right now, and the honest fix is a set of core capabilities rather than more extension-side patching. The branch carries the fixes above so it's in the best possible state if we pick it back up. Full credit to @ChonSong for the original concept (core PR #5312).

ChonSong:
Thanks for the thorough review and for committing fixes directly on the branch. I've reviewed all the changes in ef696b6 — solid work on every count.

**What's fixed and validated:**
- The `renderMd()` integration with off-origin stripping is the right approach for tile transcripts — no more raw markdown in tiles, and `network_external:false` is properly honored.
- `closeTile` detaching instead of cancelling streams is a meaningful UX improvement. The guard preventing `hideGrid()` from firing before tiling was entered also resolves the destructive path I'd missed.
- `CSS.escape` on `badge()` — can't argue with that. Same for the `#sessionList`/`.session-item[data-sid]` rewiring.
- `extension.json` added and `manifest.json` slimmed — I didn't know the gallery required the separate entry point. Noted for future submissions.

**On the architectural analysis:**

You're right. The root issue is that core WebUI is a singleton — `S`, `#msgInner`, one live stream, one load path — and the extension is faking multiplicity on top of it. Focus changes moving `#msgInner` between tiles and the 500ms poll are hacks, not a solution. Non-focused tiles are frozen snapshots, and I agree that can't be made genuinely robust from extension space.

Your proposal for A (`registerHermesSessionOpenHandler`) + B (`renderTranscript`) is the right bar for a shippable extension with the VS-Code-split model (focused tile live, others read-only). C + D/E would let every tile be live independently, which is a different product.

**Next step I can offer:**

I'm happy to contribute those core hooks (A + B) as a core PR against `nesquena/hermes-webui`. They're well-scoped, backward-compatible additions — a cancelable session-open event and a transcript renderer that accepts an arbitrary container — that don't change any existing behavior. Once those land, this extension can be updated to use the sanctioned API instead of DOM hacking, and the architectural concern is resolved for the honest use case.

If you're open to that approach, I'll draft and submit the core hooks PR and then follow up with a revision of this extension that consumes them.

Otherwise, happy to leave this parked — the branch is in the best possible state if the core hooks land later.

ChonSong:
Following up on the architecture discussion — I've submitted the two core hooks as a standalone PR against `nesquena/hermes-webui`:

**PR: https://github.com/nesquena/hermes-webui/pull/5508 `feat(core): add registerHermesSessionOpenHandler + renderTranscript for extensions`**

### Hook A — `registerHermesSessionOpenHandler(fn)`
Registers a handler called on session open. The handler receives `(sid, data, {preload, loaded, opts})`. With `{preload:true}`, returning `{cancel:true}` aborts the navigation (useful for redirecting). With `{loaded:true}`, the session is fully loaded with data.

This replaces the need for MutationObserver-based sidebar click interception. An extension just registers a handler — it fires at the right time, with the right data, regardless of how the user navigated (click, keyboard, sidebar lineage, etc.).

### Hook B — `renderTranscript(container, messages, opts)`
Renders an array of messages into an arbitrary DOM container using the core's `renderMd()` pipeline. Creates `.msg-row[data-role]> .msg-body > .msg-body-inner` matching the core transcript structure. Skips tool messages. Returns the container for chaining.

This gives extensions a sanctioned way to render message transcripts without touching `S` or `#msgInner`.

### Next steps
Once these hooks land in `nesquena/hermes-webui:master`, the chat-tiling extension can be updated to be a clean consumer:
- `registerHermesSessionOpenHandler` → no more sidebar click interception
- `renderTranscript` → proper markdown in tiles

I've already prepared this update on the `feat/chat-tiling-extension` branch — the extension checks for the hooks at runtime and uses them when available, falling back to the old approach otherwise.

ChonSong:
@nesquena Would be power feature might not be clean as is but also not sure webui hooks minimal approach 
{A413EEDE-BD2B-4366-B3AE-D04450B12E45}

nesquena-hermes:
Thanks @ChonSong. My read matches yours: as a standalone DOM-driven extension this works but stays fragile at the edges (the focus-move of `#msgInner` and the poll are the fragile parts), and the genuinely-robust version is the one that consumes the two core hooks you proposed in nesquena/hermes-webui#5508 (`registerHermesSessionOpenHandler` + `renderTranscript`).

Since the branch already has the runtime feature-detect (uses the hooks when present, falls back otherwise), the clean sequence is to let #5508 land first, then this extension ships as a proper consumer rather than baking in the DOM-hacking path as the primary one. I'd rather not merge the fallback-as-default and then immediately rewrite it.

So I'm keeping this parked on that dependency, not on anything wrong with the branch — it's in a good state. Once #5508 is merged I'll come back, verify the hook-consuming path end-to-end, and take it the rest of the way. Appreciate you doing the core-hooks PR the right way.

nesquena-hermes:
Closing this maintainer-authored copy as superseded by #60, so there is one authoritative Chat Tiling review lane.

I compared the exact heads before closing:

- #42 is the older, conflicting maintainer copy (`ef696b6`) with the four runtime entry files plus the old catalog bullet.
- #60 carries newer versions of every Chat Tiling runtime file at `cc2f70f`, the narrowed snapshot-only state model, the stale-activation guard, and the repository-level behavior harness/CI work requested during review.
- #60 is by the original contributor and preserves the right attribution/rework history.

This is a deduplication, not a claim that #60 is merge-ready. Its current `CHANGES_REQUESTED` review remains active until the rapid A→B test proves B owns the live state before stale A rejects, the hook test invokes the extension's registered handler, and CI uses `npm ci` with the committed lockfile. All further review should continue on #60.

PR #47, opened 2026-07-05
feat(theme-creator): blur intensity slider, code cleanup, and screenshot

ChonSong:
Adds a configurable **Blur intensity slider** (0–50px) to the Theme Creator, refactors duplicated slider code into reusable helpers, and cleans up stale documentation.

### Changes

- **Blur intensity slider** — controls backdrop-filter on glass surfaces (0-50px). Saved per theme, reflected in live preview.
- **Code consolidation** — extracted sliderSection() and bindSlider() helpers. No net LOC change.
- **Single style element** — merged hwxThemeCreatorCodeStyles into hwxThemeCreatorBgStyles. Removed dead CODE_STYLE_ID constant.
- **Accessibility** — both range sliders now have aria-label attributes.
- **Documentation** — README updated for blur slider, dynamic opacity, single style element, stale fixed-value references.
- **Screenshot** — added panel screenshot and wired into extension.json.
{F104020A-BD53-47D8-BBC7-2299D239B752}


nesquena-hermes:
## Full extension QA gate — theme-creator (blur slider + refactor) ✅

Thanks @ChonSong — the refactor is clean (nice net-negative LOC, the `sliderSection()`/`bindSlider()` extraction reads well, aria-labels on both sliders). I ran the full pipeline (my browser drive on an isolated server + a Codex security/breakage gate) and pushed fixes directly to the branch. All green now; flagging for one last human look since a security fix landed.

### Browser QA — the blur slider works end-to-end

Seeded a glass theme with a background image, drove the blur slider, and confirmed `backdrop-filter: blur(Npx)` genuinely applies to the live surfaces (sidebar/main/composer), scoped to the theme's `data-skin` under both `:root[data-skin]` and `:root.dark[data-skin]`. Good dual-prefix handling.

### What the Codex gate caught (all fixed + re-verified live)

1. **[SECURITY]** `bgImage` was only checked with `startsWith('data:image/')` before being interpolated into `background-image:url("…")` — a tampered `localStorage` theme record could break out of the CSS string and inject an external `url()`. Now validated on **read** against a strict anchored base64 raster data-URL grammar (`DATA_IMAGE_RE`: png/jpeg/jpg/gif/webp) + 500KB cap; invalid images are dropped. **Verified live:** a CSS-breakout payload carrying an external URL is rejected — no rule emitted, no external URL in the stylesheet. This also makes the README's "validated base64 data URLs only" claim actually true.
2. **[BREAK]** blur `0` was coerced to `20` via `|| 20` / `> 0 ?` (JS falsy) — the advertised 0–50px slider could never reach clear-glass. Fixed with `Number.isFinite` guards. **Verified live:** blur=0 now renders `blur(0px)`, not `blur(20px)`.
3. **[DISCLOSURE]** removed an undeclared `hwx-tc-debug` localStorage write.
4. **[BREAK]** `slugify()` could yield `custom-` (empty suffix) for punctuation/emoji-only names — saved+applied in-session but rejected by `validTheme()` on reload. `slugify()` now always falls back to a valid suffix, and `saveCurrent()` rejects any generated key that fails `THEME_KEY_RE` before saving.

### Verified

Static gates pass (13 entries), blur applies end-to-end incl. the 0px floor, malicious image rejected on read, and the **Codex re-gate on the fixed head returned SAFE TO SHIP with zero findings** (no regex bypass, no remaining coercion path).

@franksong2702 @rodboev — one last review welcome before merge, since a security fix (bgImage validation) landed. Nothing outstanding on my end.

Co-authored the fix commit crediting @ChonSong.

PR #51, opened 2026-07-07
feat: Docker & Tunnel Manager extension

ChonSong:
## Docker & Tunnel Manager — Hermes WebUI Extension

Manage Docker containers, images, volumes, compose projects, and Cloudflare tunnels from inside the Hermes WebUI. 7-tab panel with live stats, health checks, log streaming, and inline actions.

---

### 🔧 Sidecar (`sidecar/docker-tunnel-sidecar.py`)

Single-file Python HTTP server (port 17900, bound to `127.0.0.1` only) that powers all tabs:

| Endpoint | Tab | Description |
|---|---|---|
| `GET /api/containers` | Containers | List all containers + live CPU/mem stats (via `docker stats --no-stream`) |
| `POST /api/containers/{id}/{action}` | Containers | Start / stop / restart containers |
| `POST /api/containers/prune` | Containers | Prune stopped containers |
| `GET /api/images` | Images | List images with size + container refcount |
| `POST /api/images/prune` | Images | Prune dangling images |
| `GET /api/images/{id}/history` | Images | Image layer history |
| `GET /api/system/df` | System | Docker disk usage with reclaimable breakdown |
| `GET /api/tunnels` | Tunnel | Cloudflare tunnel info (ID, connectors, parsed ingress rules) |
| `GET /api/tunnels/health` | Tunnel | Parallel ingress health checks (status/HTTP code/latency) |
| `GET /api/tunnels/logs` | Tunnel | Recent tunnel logs via journalctl |
| `GET /api/volumes` | Volumes | List volumes with driver, mountpoint, size, linked containers |
| `POST /api/volumes/prune` | Volumes | Prune unused local volumes |
| `POST /api/volumes/{name}/delete` | Volumes | Delete a specific volume |
| `GET /api/compose` | Compose | Discover Compose projects via Docker labels; shows running/total counts and services |
| `GET /api/containers/{id}/logs?tail=N&follow=true` | Logs | SSE (EventSource) live log streaming |

---

### 🧩 Frontend (`assets/docker-tunnel-manager.js`)

IIFE-pattern extension, no framework deps. Pure DOM injection, scoped CSS.

- **Rail button** — Lucide-style inline SVG (box with tunnel arch), inserts beneath Theme Creator
- **7-tab overlay panel**: Containers (live CPU/mem), Images, System (disk usage bars), Tunnel (health checks + logs), Volumes (prune/delete), Compose (project cards), Logs (SSE streaming)
- Toast notifications, confirmation dialogs, auto-refresh with configurable interval

---

### 📐 CSS (`assets/docker-tunnel-manager.css`)

~410 lines, scoped under `.hwx-dtm-*`, inherits WebUI CSS variables. Honour `prefers-reduced-motion`.

### ⚙️ Manifest & Proxy

- `manifest.json`: `loopback-sidecar` declaration on `127.0.0.1:17900`
- WebUI proxies all `fetch()` through `/api/extensions/docker-tunnel-manager/sidecar/` — works through Cloudflare tunnels

### 🧹 Code reductions

- `tunnel_adapter.py` → inlined into the sidecar (Ngrok/Bore stubs dropped → fewer files, less code)
- Compose inline styles → CSS classes
- Stripped unused CSS rules and localStorage fallback
- Total: **-392 lines, -1 file** with no functionality lost

---

### 🚀 Quick Start

1. **Sidecar systemd service** — copy from repository, then:
```
systemctl --user daemon-reload
systemctl --user enable --now hermes-docker-tunnel-sidecar.service
```
2. **Register extension** — add to `extension-install-manifest.json`:
```json
{ "enable_extension_ids": ["docker-tunnel-manager"] }
```
3. **Proxy consent** — add to `extension-overrides.json`:
```json
{ "sidecar_proxy_consents": { "docker-tunnel-manager": "http://127.0.0.1:17900" } }
```
4. **Restart WebUI** and refresh.

---

### ✅ Verified

- All 15 API endpoints tested (sub-second responses)
- Container start/stop/restart and prune confirmed working
- Volume list/prune/delete working
- Compose project discovery (4 projects, 17 containers)
- Cloudflare tunnel info parsing + 22 ingress routes health-checked
- SSE log streaming through EventSource
- WebUI proxy integration through Cloudflare Access
- Sidecar stats via `docker stats --no-stream` (avoids docker-py hang)
- Persistent systemd service (auto-starts on boot, auto-restarts on failure)


nesquena-hermes:
## New sidecar standard applies to this PR

@ChonSong, the shared sidecar authentication work has now merged:

- Library scaffold, contract, and CI: [hermes-webui/hermes-webui-extensions#65](https://github.com/hermes-webui/hermes-webui-extensions/pull/65)
- Authoritative contract: [`docs/SIDECAR_CONTRACT.md`](https://github.com/hermes-webui/hermes-webui-extensions/blob/main/docs/SIDECAR_CONTRACT.md)

This PR also needs that standard because its sidecar exposes Docker/container/volume mutations and host/tunnel data. Binding to `127.0.0.1` and relying on CORS does not authenticate direct loopback callers.

Please rebase onto current `main`, then choose one of the contract's two runtime models:

### Recommended here: external token-v1 runtime

The current sidecar depends on installed `docker-py` and provides SSE logs. A repository-vendored runtime must run the canonical `/usr/bin/python3 -S -u sidecar.py`; `-S` hides installed site packages, and the WebUI proxy does not support SSE. For this dependency-heavy host-control service, the cleaner fit is:

```json
"sidecar": {
  "type": "loopback",
  "origin": "http://127.0.0.1:17900",
  "health_path": "/health",
  "proxy_auth": "token-v1",
  "runtime": {
    "kind": "external",
    "repository": "https://github.com/ChonSong/<runtime-repo>"
  }
}
```

Repeat `proxy_auth: token-v1` in `manifest.json`, but omit `runtime` there. Move the server implementation to the declared repository and implement the language-neutral contract:

- Read the per-extension token using the documented resolution order (`HERMES_EXT_SIDECAR_TOKEN_FILE`, then the WebUI state dir/Hermes home/platform default).
- Require `X-Hermes-Sidecar-Token` on every route except `GET/HEAD /health`.
- Return 401 for missing/wrong token, 503 when the token file is absent/unreadable, use constant-time comparison, and re-read the file per request so rotation is live.
- Keep `/health` liveness-only with `Access-Control-Allow-Origin: *` and `Cache-Control: no-store`. The current `/api/health` path must become `/health` for scaffold/contract parity.
- Do not rely on browser `Origin`/CORS as authentication.

The browser should use only `/api/extensions/docker-tunnel-manager/sidecar/...`, after explicit consent. With WebUI authentication off, token-v1 returns 403 by design; the operator must first enable **Settings → Password**.

The current EventSource log stream must also change. The proxy buffers responses and times out around 10 seconds, so expose bounded log snapshots or start-job + poll rather than `follow=true` SSE through the proxy.

### Alternative: keep it vendored

If you keep the runtime inside this repository, it must use the byte-identical files in [`examples/sidecar-scaffold/`](https://github.com/hermes-webui/hermes-webui-extensions/tree/main/examples/sidecar-scaffold), declare `runtime.kind: vendored`, move routes into `routes_impl.py`, remove every custom `HTTPServer`, and either remove or vendor the `docker-py` dependency so the required `-S` launch works. CI then requires:

```bash
```

Core #6331 first appears in `exp-v0.52.129`; until stable promotion, document the requirement as “a WebUI build containing #6331.” This comment covers the new sidecar-contract delta only; it does not settle the separate product-scope and host-control review for this entry.

nesquena-hermes (changes requested):
## Full extension QA — kick back on `435e239d8473df6200cdbd717286b819b92e3fea`

I reviewed all five extension files plus the current external runtime at `ChonSong/docker-tunnel-sidecar@a3cd1f82910261ab386f10fd1a69e2c6717b5bf7`, ran the repository gates, and completed Codex/Fable passes. This needs coordinated work in both repositories before another review; it is not close enough for a maintainer patch.

### Blocking correctness and trust issues

1. **Authenticated-origin XSS** (`assets/docker-tunnel-manager.js:101,349`). `t()` escapes only angle brackets, but sidecar-controlled `mountpoint` is interpolated into a quoted `title` attribute. A quote breaks out of the attribute. Build response-derived attributes with DOM APIs or use a context-safe encoder covering `& < > " '` everywhere.
2. **Permission disclosure is materially false** (`extension.json:44-63`). `webui_api.read/write` are empty despite sensitive sidecar reads and destructive container/volume operations. `network_external:false` is also false: the runtime performs outbound ingress checks and Cloudflare CLI operations. Enumerate every proxied read/write route and declare/disclose external network activity.
3. **The Docker-authority runtime is mutable and unpinned** (`extension.json:27`; `README.md:34`). Installation clones the default branch without a release, commit, or checksum. Vendor the reviewed runtime, or pin an immutable release/commit plus checksum and make the review artifact reproducible.
4. **Panel lifecycle is broken** (`assets/docker-tunnel-manager.js:484,564,641`). `stopPolling()` is local to `renderLogs`, but outer `switchTab` and `closeOverlay` call it. Leaving Logs or closing the panel throws and can leave polling alive. Move cancellation to IIFE scope and invalidate in-flight polls so detached work cannot reschedule.
5. **Several live routes never match, while malformed prune paths do** (external `sidecar.py:309-340`). Container logs, image history, and volume deletion compare a three-element slice to a two-element list and always 404. Conversely, paths such as `/api/containers/prune/extra` still invoke destructive prune. Match every normalized path and segment count exactly, URL-decode IDs/names, and remove or deliberately expose the undocumented system-prune route.
6. **The documented service cannot start as written** (`README.md:34`; external `sidecar.service:18-20`). The clone path and `WorkingDirectory` disagree; `python3 -S` prevents the documented site-packages `docker-py` install from importing; the manifest docs URL points to `blob/main` while the runtime branch is `master`. Use one real install path and a pinned repository-local venv.
7. **Author-specific tunnel defaults leak into a general entry** (external `sidecar.py:77,210`). `codeovertcp` and `gto-wizard-tunnel.service` are hard-coded. Derive the tunnel from config and make the validated journal unit/config path explicit settings; reject nonzero CLI results.
8. **Proxy-envelope bounds are not honored** (external `sidecar.py:104,156,304`). Several operations exceed the ~10 s proxy deadline; `tail`/`lines` are unbounded; responses have no encoded-byte ceiling. Use start-job/poll for long work, enforce an overall deadline, clamp counts, and cap/paginate responses below 512 KiB.
9. **The sidecar contract claim does not match implementation** (external `sidecar.py:274`). Unauthenticated `OPTIONS` is accepted on every route, `HEAD /health` is not implemented, and a non-ASCII invalid token can close the connection. Route all methods through one auth/exception chokepoint and support bodyless GET/HEAD health only.
10. **Normal Docker responses can become false success or 500s** (external `sidecar.py:416,458,475,497`). `*Deleted: null` reaches `len(None)`, nullable disk-usage collections crash, system-prune reports success even if every prune failed, Compose failures look like zero projects, and volume usage is read from the wrong source. Normalize nullable fields and return explicit non-2xx/partial-failure results instead of optimistic empties.

### Product/UX truthfulness

- `sidecarPort` is dead; the proxy origin is manifest-controlled and JS never reads the setting. Remove it or implement a core-supported reconfiguration flow.
- The README advertises prune controls, image history, detailed Compose management, configurable auto-scroll/line count, and “No data leaves the machine,” but the current UI does not implement those controls and the runtime performs outbound checks. Implement and prove each claim or remove it.
- The confirmation dialog escapes its own `<strong>` markup, so the action name is not emphasized. Build the dialog with text nodes/DOM rather than mixing escaped data and trusted markup.

Please add automated coverage in the external runtime for token/auth routing, GET+HEAD health, bounds/deadlines, destructive route behavior, and the actual systemd/venv install path. Re-review should wait until both repositories are pinned to matching heads and the metadata, README, runtime, and UI describe the same authority and behavior.

ChonSong:
## Closing PR #51 — Starting fresh on Docker & Tunnel Manager

After the full QA review by @nesquena-hermes (`4814151049`), this branch has 10 blocking issues across both the extension files and the external `docker-tunnel-sidecar` runtime — including XSS, false permission disclosure, an unpinned supply chain, broken route matching, and an unreachable service setup. The reviewer's verdict was "not close enough for a maintainer patch."

Fixing this branch in-place would require coordinated rewrites in both repos while carrying poisoned metadata in the commit history. Cleaner approach:

1. **Fix the sidecar** — pin a reviewed release with correct routing, auth contract, bounds, and no hardcoded tunnel names
2. **New branch off `hermes-webui:main`** — correct permissions, clean lifecycle, honest README, pinned to the fixed sidecar
3. **Re-request review** from the same reviewer with both heads matched

The 7-tab panel design and ~400 lines of scoped CSS are solid and will carry forward into the new PR.

---
*Migrated to new approach — see linked PR when opened.*

PR #49, opened 2026-07-06
feat: add voice-control-session-management extension

ChonSong:
## Session Orchestrator — Voice-Commanded Multi-Session Management

A new extension that lets users create, switch, prompt, read, and monitor
concurrent agent sessions using voice commands or text shortcuts.

### Thinking Path

The existing voice mode (boot.js) provides SpeechRecognition + TTS but routes
everything to the composer textarea. The chat-tiling extension provides a
multi-session grid but no voice integration. The SessionChannel SSE delivers
background-task-complete events but has no consumer outside `messages.js`.

This extension sits at the intersection: it hijacks `_voiceModeSend` to parse
commands before they reach the send pipeline, uses `registerAlias`/`resolveAlias`
to map friendly names to session IDs, and wires into the existing notification
system (`playAttentionSound`, `sendBrowserNotification`, `_handleBgTaskCompleteEvent`)
for async completion alerts.

### What Changed

```
extensions/session-orchestrator/
├── assets/orchestrator.js   — full extension logic (~530 lines)
├── extension.json           — gallery metadata with permissions
├── manifest.json            — runtime manifest + settings_schema
└── README.md                — install, usage, trust model, limitations
```

### Key Decisions

1. **No new backend routes.** Everything rides on existing infrastructure:
   `window.newSession()`, `_voiceModeSend`, `_handleBgTaskCompleteEvent`,
   `playAttentionSound`, `sendBrowserNotification`.
2. **Command parser is a pure function** (~20 lines, object schema + regex
   matching). No FSM library dependency — the existing voice mode already has
   the state machine (`_voiceModeState: idle|listening|thinking|speaking`).
3. **Alias registry persists to localStorage** (`hwx-orch-state` key). Survives
   page refreshes and tab crashes.
4. **Phase 1 is extension-only.** Phase 2 (containerized desktop environments)
   would use the hermes-desktop pattern and is explicitly called out as a
   limitation in the README.

### Why It Matters

- Solves the "I want to start a research session while waiting for the code
  review session to finish" workflow without tab-juggling
- Voice-first input means users can manage sessions while keeping hands on
  keyboard for actual work (or while away from the desk)
- Composable with chat-tiling (badges on sidebar rows, tile maximize)

### Verification

- `node scripts/validate-extensions.mjs` — passes
- `node scripts/scan-extension-safety.mjs` — passes (orchestrator section clean)
- `node --check extensions/session-orchestrator/assets/orchestrator.js` — syntax ok

### Risks / Follow-Ups

[...]

PR #53, opened 2026-07-07
feat: add Fun Audio Chat Connector extension

ChonSong:
## Fun Audio Chat Connector

Full-duplex speech-to-speech voice chatting via a local [Fun-Audio-Chat (FAC)](https://github.com/FunAudioLLM/Fun-Audio-Chat) server — Alibaba's 8B Large Audio Language Model (Apache 2.0).

### What it does

Connects the WebUI directly to a local FAC WebSocket server (`ws://127.0.0.1:11236/chat`) for real-time voice conversations:

- **Mic capture**: `getUserMedia` → `MediaRecorder(Opus)` → binary WebSocket frames
- **Streaming playback**: received Opus frames → `AudioContext` → speakers
- **FAC binary protocol**: `0x00` handshake, `0x01` audio/Opus, `0x02` text, `0x03` control
- **Overlay panel**: floating mic button bottom-right, collapsible voice panel
- **VU meter**: real-time audio level via `AnalyserNode`
- **Settings**: host/port, mode (S2S/S2T/TTS-only), persisted via extension storage

### Architecture

```
Browser (WebUI)
  getUserMedia → MediaRecorder(Opus) → WebSocket ──→ FAC server (:11236)
  AudioContext ←── Opus audio stream ←──
```

No core WebUI changes, no sidecar, no backend routes. Connects directly to FAC via loopback WebSocket (already allowed by default CSP).

### Permissions

- `dom.owned: true` — creates its own overlay DOM
- `storage.owned: true` — saves settings (host, port, mode)
- `network_external: false` — FAC runs on localhost, CSP already allows `ws://127.0.0.1:*`
- Microphone via browser `getUserMedia` (user prompt at runtime)

### Background

This is part of a broader FAC integration effort. A companion Hermes Agent plugin (`fac-s2s`) that registers FAC as a server-side TTS/STT provider for all gateway clients is published separately as a standalone plugin repo per CONTRIBUTING.md policy.

### Status

- [x] Full binary protocol implementation (handshake, audio, text, control frames)
- [x] Mic capture via MediaRecorder(Opus) — same codec as FAC expects
- [x] Streaming audio playback via AudioContext
- [x] Floating mic button + collapsible overlay panel
- [x] VU meter with AnalyserNode
- [x] Settings persistence (host, port, mode)
- [x] Connection status indicator (disconnected/connecting/connected/error)
- [x] Transcription log in panel
- [x] Zero core WebUI changes
- [x] Validated: JSON parses clean, SVG valid

## Cross-Extension Event Integration

The connector now publishes to the shared CustomEvents bus so sibling extensions
(emotion-avatar #56, session-orchestrator #49) can react without tight coupling.

### Dispatches
- `hermes:voice:state` — `{ state: 'listening' | 'idle' }` on connect/disconnect
  and on record start/stop. Derived from real connection + recording state.
- `hermes:fac:emotion` — `{ emotion, confidence, source }` emitted when the ASR
  transcript matches a lightweight client-side lexicon
  (happy / sad / angry / surprised / confused).

### Note on emotion
The FAC server protocol does **not** currently return a structured emotion field,
so emotion is inferred locally from the transcript (non-fatal, swappable). Replace
`emitEmotionFromText` with server emotion frames if the protocol adds an emotion
channel — no other call sites change.

### Listens
- Nothing yet (pure producer). It does not depend on any other extension being
  installed; the avatar degrades gracefully if emotion-avatar is absent.


nesquena-hermes:
## Full extension QA gate — fun-audio-chat-connector (draft)

Thanks @ChonSong — fun concept, and the binary-protocol / Opus streaming plumbing is nicely structured. I ran the complete extension QA pipeline on this (my own browser drive on an isolated server + two rounds of the Codex security/breakage gate) and pushed fixes directly to the branch. It's a draft, so I've **left it in draft** — one architectural item on the audio path is genuinely yours to finish + verify against a running FAC server (details at the bottom). Everything else is fixed and verified.

### What I fixed and verified (pushed to the branch)

**Security / honesty**
1. **`network_external: false` was not actually enforced.** `getWsUrl()` built `ws://${host}:${port}/chat` straight from user-settable inputs with no validation — a typed (or tampered-localStorage) host could stream **microphone audio to any external WebSocket**. And the WebUI CSP is *report-only* in some deployments, so it couldn't be relied on to block it. Added `isLoopbackHost()` (localhost / 127.0.0.0/8 / ::1 only) + `sanitizePort()`, enforced at **both** save-time (reject + reset to 127.0.0.1 with a log line) and in `loadSettings()` (defensive read). **Verified live:** a tampered `evil.example.com` is sanitized back to `127.0.0.1`. (Also removed `0.0.0.0` from the allow-list on the re-gate — it's the all-interfaces bind address, not loopback.)

**Broken behavior**
2. **The core feature didn't work — audio was silently dropped.** `MediaRecorder.ondataavailable` emits `Blob`s, but `sendFrame()` had no `Blob` branch and just `return`ed — so every audio frame was discarded and nothing ever reached FAC, even though the UI showed "recording." Made the handler async and convert via `event.data.arrayBuffer()` before `sendFrame`, with a post-await connection re-check.
3. **No usable way to connect.** On a fresh panel, `init()` never called `setStatus()`, so the Talk button stayed disabled *and* the Reconnect button was never revealed (its `.hwx-fac-show` class is only toggled inside `setStatus()`) — a dead-end. `init()` now drives `setStatus(DISCONNECTED)`. **Verified live:** Reconnect shows on load, Talk correctly disabled until connected.

**Hygiene / leaks**
4. **Deleted `extension.html`** — a 548-line standalone duplicate, unreferenced by `manifest.json`, not consumed by the registry, that carried the *same* unvalidated-`ws://` + dropped-`Blob` defects and undeclared storage keys. Removed the file + its `assets.pages` entry + the README row.
5. **Unbounded log growth** → capped at 300 entries (a long session / reconnect loop grew the DOM without limit).
6. **AudioContext leak** — `disconnect()` never closed it (browsers cap ~6 concurrent), so a connect/disconnect cycle leaked one each time. Now closed + nulled.
7. **README** — the `network_external` note now accurately describes the in-code loopback enforcement and the separate-local-app FAC model (no bundled sidecar, no backend route).

### One item left for you (why it's not merged)

**Audio chunk ordering + `end_turn` race** (`connector.js` `ondataavailable` / `stopRecording`). My Blob→ArrayBuffer fix makes audio *send*, but the conversions are async and not serialized: chunks can be sent out of order, the final chunk can arrive *after* the synchronous `end_turn`, and a chunk from an old socket can land on a freshly-reconnected one. The correct fix is a **per-recording send queue** bound to the current recorder/session/WebSocket, with `end_turn` sent from `mediaRecorder.onstop` only after that queue drains and the same socket is still open.

I stopped there deliberately: it's an architectural change to the core audio path, and I **can't verify it end-to-end without a running FAC server** — I have no way to confirm the Opus frame format is what FAC expects or that ordering actually matters to its protocol. That's yours to implement and test against a live FAC instance. Once it's in and you've confirmed a real voice round-trip works, mark the PR ready and I'll re-gate + merge.

### On the permission model

Codex suggested setting `loopback_sidecar: true`. I left it `false` on purpose: that capability means the extension *ships/manages* a sidecar the loader proxies — FAC is a **separate app you run independently**, reached by a direct browser→loopback WebSocket. The honest disclosure is the README note (added), not a `loopback_sidecar` claim. Happy to revisit if you disagree.

Static gates pass; the security/hygiene fixes are verified. Co-authored the fix commits crediting you.

ChonSong:
no longer planned

PR #58, opened 2026-07-11
Use real WebUI CSS variables so extensions render in the light theme

ai-ag2026:
## Problem

Several extensions reference CSS custom properties the WebUI **never defines** (e.g. `--text-primary`, `--bg-elevated`, `--danger`). Because those variables are absent, `var()` always falls through to the hardcoded fallback — which is a dark-mode color. The result: these extensions render dark-only, and under the **light theme** their text/surfaces come out wrong (e.g. light text on a light background).

The real variables defined in `static/style.css` are `--text`, `--muted`, `--surface`, `--surface-subtle`, `--error`, `--accent-text`, etc.

## Fix

Replace the nonexistent variable names with the real ones (fallbacks kept, so nothing regresses if a variable is ever missing):

| used (undefined) | real style.css var |
|---|---|
| `--text-primary` | `--text` |
| `--text-secondary` | `--muted` |
| `--bg-elevated` | `--surface` |
| `--danger` | `--error` |
| `--surface2` | `--surface-subtle` |
| `--text2` | `--muted` |
| `--accent-contrast` | `--accent-text` |

Affected extensions: **custom-avatar, message-pins, theme-creator, external-app-tab, mcp-tool-shortcuts**. No behavior change beyond correct theming.

## Test

- `node scripts/validate-extensions.mjs` → all five pass.
- `node scripts/scan-extension-safety.mjs` → pass.
- Cross-checked against `static/style.css`: 0 undefined `var(--…)` references remain in the touched files.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/extensions/custom-avatar/assets/custom-avatar.css
+++ b/extensions/custom-avatar/assets/custom-avatar.css
@@ -51,7 +51,7 @@
   border-color: var(--border, rgba(127, 127, 127, .4));
 }
 .hwx-avatar-btn--clear {
-  color: var(--danger, #e5534b);
+  color: var(--error, #e5534b);
 }
 .hwx-avatar-status {
   font-size: 11px;
--- a/extensions/external-app-tab/assets/external-app-tab.css
+++ b/extensions/external-app-tab/assets/external-app-tab.css
@@ -132,7 +132,7 @@
 }
 .hwx-extapp-input:focus { outline: none; border-color: var(--accent, #6366f1); }
 .hwx-extapp-config-note { font-size: 11px; color: var(--muted, #888); line-height: 1.5; }
-.hwx-extapp-config-err { font-size: 12px; color: var(--danger, #e5534b); }
+.hwx-extapp-config-err { font-size: 12px; color: var(--error, #e5534b); }
 .hwx-extapp-config-actions { display: flex; justify-content: flex-end; gap: 8px; }
 
 @media (prefers-reduced-motion: reduce) {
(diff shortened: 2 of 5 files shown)


nesquena-hermes:
## Reviewed + merging — thanks @ai-ag2026 👍

Solid, well-targeted fix. I ran it through the full extension QA (validator, safety scan, registry generation, plus the Codex security/breakage gate) and verified every variable rename against the live core `static/style.css`.

**Verified correct:** the renamed tokens (`--danger`→`--error`, `--text2`→`--muted`, `--surface2`→`--surface-subtle`, `--accent-contrast`→`--accent-text` where it's link/normal-background text) all move from names core never assigns (so they always fell through to a dark hardcoded fallback) to names that carry real per-theme values (`--error` has 23 theme definitions, `--muted` 27, `--accent-text` 39, `--surface-subtle` 24). So this genuinely fixes the light-theme rendering. Your "0 undefined `var(--…)` references remain" claim checks out.

**One spot I adjusted before merge** (message-pins, 2 lines): the active pin button renamed `--accent-contrast`→`--accent-text`, but that button sits on a solid `var(--accent)` background. `--accent-text` is accent-*colored* text intended for a normal background — in most themes it's equal or near-equal to `--accent` itself (github: both `#0969DA`; catppuccin: both `#8839EF`), so on an accent fill it renders accent-on-accent and the icon effectively disappears. The right token for text *on* accent is `--accent-contrast` (theme-creator generates it via `readableOn(accent)`, and boot.js allows it as a skin token). I reverted just those two lines to `var(--accent-contrast, #fff)` — a net no-op vs main for message-pins, keeping the correct fixes in the other four files. Pushed that as a follow-up commit on your branch with you as co-author.

Also worth a follow-up (not in this PR's scope): `assistant-avatar` has the same bug class (`--bg2`/`--bg3`, both undefined). I'll track that separately.

Merging now. Appreciate the clean, precisely-scoped contribution.

PR #64, opened 2026-07-19
Add rss-feeds extension

asorourx:
Adds the **rss-feeds** extension: a full RSS/Atom reader for the WebUI — a
two-pane overlay with categorized subscriptions, keyword filters, full-text
search, cross-device read tracking, and optional free/local AI summaries
(local Ollama / OpenRouter `:free` / Gemini free), all via a `127.0.0.1`
loopback sidecar. `manifest-bundle` + `loopback-sidecar`; no core views modified
beyond a titlebar launcher button.

> **Split note:** this PR originally bundled `profile-avatars` and `sysinfo`.
> Per review it's been narrowed to **rss-feeds only**; the other two are now on
> their own PRs so this one can land independently.

### Review fixes applied (incremental, on this branch)
- **Auto-consent removed** — sidecar proxy no longer self-granted; UI reads
  `/api/extensions/status` and shows an "approve in Settings" state.
- **SSRF hardening** — resolve → reject any non-global IP (loopback/private/
  link-local/CGNAT/metadata) → pin to the validated IP, re-pinned each redirect
  hop, on both feed fetch and article retrieval.
- **XSS** — quote-safe `esc()`, http(s)-only link rewriting, delegated listeners
  (no inline `onclick`).
- **Credential-leak on redirect** — redirects disabled on the credentialed LLM
  calls; Gemini key moved to the `x-goog-api-key` header.
- **SSH tunnel path removed** — keeps rss-feeds a clean loopback sidecar,
  `native_host:false` honest.
- **Disclosure** — README enumerates every sidecar egress + filesystem read;
  `network_external:true`; all `mc.feeds.*` storage keys declared.
- **`/health` CORS** — `ACAO:*` + `Cache-Control:no-store`, scoped to `/health`.

### Known dependency
The shared proxy→sidecar token boundary is a **core-side** change (being defined
by the maintainer). Sidecar-side token validation is deferred until that
contract lands; everything else is complete.


asorourx:
Screenshot 2026-07-13 at 3 38 08 PM
Screenshot 2026-07-13 at 3 30 00 PM
Screenshot 2026-07-13 at 3 30 16 PM
Screenshot 2026-07-13 at 3 30 34 PM
Screenshot 2026-07-13 at 3 30 52 PM
Screenshot 2026-07-13 at 3 31 50 PM
Screenshot 2026-07-13 at 3 32 03 PM
Screenshot 2026-07-13 at 3 32 27 PM
Screenshot 2026-07-13 at 3 32 42 PM
Screenshot 2026-07-13 at 3 32 59 PM

nesquena-hermes:
## Review — `rss-feeds` (1 of 3)

Thanks for this, @asorourx — genuinely high-craft work, and the first end-to-end `loopback-sidecar` reference in the registry. I reviewed it through a security gate (Codex, full repo context, live-verified) and a UX gate (Fable), plus my own read. Verdict for this one: **needs fixes before merge (no merge-blocker that isn't fixable on-branch).**

### 🔴 Cross-cutting critical (applies to all three extensions in this PR — see the shared note at the bottom of the sysinfo comment)
The sidecar sets auth off and the frontend auto-grants its own proxy consent. Details + the shared fix are in the sysinfo comment so I don't repeat them three times — but they apply here too (`sidecar.py`, `shim.py:45-50`, `feeds-inject.js:40-45,138`).

### 🔴 Security (must fix)
1. **SSRF guard is bypassable (`feeds.py:1133`).** It resolves the host once for validation, then `urllib` resolves *again* to connect — a DNS-rebinding gap — and the predicate permits non-global space like `100.64.0.0/10` incl. the Alibaba metadata IP `100.100.100.200` (Codex verified). Fix: reject any address where `ip.is_global` is false, then **connect to the validated IP** while preserving the Host header + HTTPS SNI/cert validation, and re-resolve+re-pin on **every** redirect hop with a bounded hop count. Apply to both `_http_get()` and `_fetch_article_text()`.
2. **XSS via non-quote-safe `esc()` (`feeds.js:127`).** `esc()` isn't quote-safe but is used inside quoted attributes and inline `onclick` strings for feed names, categories, keywords, model names, URLs, errors; `_miniMarkdown()` (`:2164`) drops an LLM-controlled URL straight into `href`, and feedparser preserves `javascript:` entry links stored without scheme validation (`feeds.py:1260`). Fix: build nodes with DOM APIs + `textContent`, bind listeners instead of inline handlers, and constrain all links to `http:`/`https:` at both ingestion and render.
3. **Credential leak across redirects (`feeds.py:199`).** OpenRouter calls use the default redirect handler with an `Authorization` header — Python copies that header to a cross-origin redirect, so `OPENROUTER_API_KEY` can leave the intended host. Gemini key is in the URL query (`:253`). Fix: disable redirects (or origin-lock them) for both credentialed clients; move the Gemini key to an `x-goog-api-key` header on a fixed URL.

### 🟠 Disclosure (permission-vs-code honesty — Frank reviews these line-by-line, so worth getting right)
4. **`native_host:false` is wrong (`extension.json:52`).** `_ensure_ssh_tunnel()` (`feeds.py:155`) spawns a detached `ssh -N -L … StrictHostKeyChecking=accept-new` that uses the user's SSH identities/agent + TOFU host-key enrollment and can outlive the sidecar. This is genuinely surprising in an RSS reader. Declare `native_host:true`, validate host/port, harden the invocation (`BatchMode=yes`, `-T`, `-F /dev/null`, no agent/X11 forwarding, `--`), track+terminate the child, and document it. (Honestly — consider whether the remote-ollama SSH path belongs in this extension at all.)
5. **`network_external` disclosure is inaccurate (`README.md:38`).** The README says it's "only for Share links / no background external requests," but the real outbound surface is feed fetches, article-page retrieval, DuckDuckGo/Google favicon requests, and OpenRouter/Gemini calls **carrying article content**. Rewrite the trust section to enumerate every destination, trigger, and data class; drop the "one reason"/"no background requests" claims.
6. **`filesystem.arbitrary:false` understates it (`extension.json:53`).** The sidecar reads `~/.hermes/.env` for API keys and inspects `/proc/<pid>`. Either drop those reads (accept explicit config/env) or declare + disclose the credential/process inspection.
7. **Undeclared storage keys (`extension.json:44`).** `feeds.js:968,1021` read/write `mc.feeds.pageSize`, `.pageSize.clicked`, `.pageSize.summaries` — none declared. Add them (+ the `favicons/` disk cache) and update the README.

### 🟡 Breakage
8. **`/health` needs CORS (`sidecar.py:35`).** Core's diagnostics probe does a direct cross-origin browser fetch to the loopback port; without `Access-Control-Allow-Origin: *` on `/health`, a healthy sidecar reports unreachable. Add it (+ `Cache-Control: no-store`) to `/health` only.

### 🎨 UX (Fable gate — SHIP-WITH-UX-FIXES; the panel itself is well-built)
The two-pane reader, empty/loading/error states, `navigator.share`-first sharing, and mobile thumb-cluster are all genuinely good and crown-jewel-safe. Five small fixes:
1. **Hardcoded dark status colors** (`#3fb950/#ff7b7b/#d29922/…` in `feeds.css` + `feeds.js:1511`) break the light theme — regresses the PR #58 convention. Use `var(--success/--error/--warning, …)`.
2. **Escape layering bug**: Escape closes the underlying overlay while a body-mounted settings/summarize/share popup (z-9999) stays stranded **over the chat window** (no backdrop). Close the topmost layer first; have `closeOverlay()` tear down the body-mounted popups.
3. **Add `-webkit-app-region: no-drag`** to `.hx-feeds-launcher` (titlebar is a drag region; native children opt out).
4. **Launcher tooltip**: swap native `title=` for the host's `has-tooltip has-tooltip--bottom` + `data-tooltip` (the app deliberately replaced native tooltips, #1775).
5. **Reduced-motion**: the `prefers-reduced-motion` rule (`feeds.css:1273`) is overridden by specificity by the `--busy` rule (`:1319`) so the indeterminate animation still runs. Raise its specificity / reorder.
(Non-blocking polish: dead `.app-titlebar-right` primary selector, `aria-modal`+focus trap.)

**Screenshots to attach for maintainer sign-off:** titlebar launcher (light+dark), mobile titlebar next to hamburger, the Escape repro reaching the chat window, and the mobile full-screen overlay with the thumb cluster open.

Once the security + disclosure items are addressed I'll re-run both gates on the new head and confirm they flip to clean.

nesquena-hermes:
## Review — `profile-avatars` (2 of 3)

@asorourx — this one has the cleanest *UX* of the three (Fable liked it), but the security gate found a **redesign-level auth problem**, so honest verdict: **kick back for rework before re-review.** The core issue isn't a one-line patch — it's the sidecar trust model — so I want to flag it clearly rather than nibble at symptoms.

### 🔴 Security — must fix (Codex, live-verified)
1. **Authentication is hard-coded OFF (`sidecar/shim.py:121`).** `is_auth_enabled()` returns `False`, so `_require_auth()` always allows POST/DELETE, and the list/GET routes have no guard at all. Codex verified live: **unauthenticated, cross-origin requests returned 200 for upload, read, list, and delete.** That's cross-site loopback avatar replacement + unbounded arbitrary-profile row creation. The catch that makes this a *redesign*: the core proxy strips cookies/auth headers before forwarding to the sidecar, so the stubbed "WebUI cookie" checks can't work as written — you need a real proxy-injected credential (see the shared fix in the sysinfo comment). This is the reason for the kick-back.
2. **Auto-granted consent (`assets/avatars.js:23,474`).** `grantConsent()` unconditionally posts `{approved:true}` and `init()` calls it — core wraps same-origin fetches with its CSRF token, so this silently bypasses the explicit Settings → Extensions sidecar-consent decision. Remove it; leave the extension inactive with a clear "approve in Settings" message until the user consents. (All three extensions in this PR do this — shared note in the sysinfo comment.)
3. **Stored DOM XSS (`assets/avatars.js:363-377`, also flagged at `:97`).** `p.label` — which for the default profile comes from the user-configurable `window._botName` (Hermes allows arbitrary bot-name text) — is concatenated into `meta.innerHTML`. Build `.pa-m-name`/`.pa-m-sub` as elements and assign `textContent`.

### 🟠 Breakage / disclosure
4. **1 MiB upload cap vs core's 512 KiB proxy limit (`sidecar/avatars.py:30`, `avatars.js:403`).** Files between 512 KiB and 1 MiB upload "successfully" then return proxy 502 and never render — a confusing dead end. Cap both server + client at 512 KiB (and fix the README's 1 MiB references) unless core's response cap is raised first.
5. **Under-declared API surface (`extension.json:32`).** `webui_api` lists only the consent write but omits the actual proxied avatar list/blob **reads** and POST/DELETE **writes** under `/api/extensions/profile-avatars/sidecar/api/avatars`. Declare them (and once auto-consent is removed, drop the consent endpoint from `write`).
6. **Archived-session mis-attribution (`avatars.js:196,218`).** The session map requests `limit=500` (ignored by current core) and omits `include_archived=1`, so archived rows fall through and get **falsely assigned the active profile's avatar**. Use the supported archived pagination params and don't attribute unknown session IDs to the active profile.

**Verified clean (credit where due):** SQL is parameterized; a multipart-disguised SVG was rejected 415; oversized input rejected 413; multipart parsing is raw-size-bounded with no archive decompression; blobs are served only as PNG/JPEG/WebP with `X-Content-Type-Options: nosniff` via `` blob URLs (no HTML/SVG execution path); selectors + re-render observers match current core.

### 🎨 UX (Fable — SHIP-WITH-UX-FIXES; the visual integration is the best of the three)
The re-skin approach (in-place on native role badges/chips rather than new controls) is exactly the right shape for a crown-jewel surface, and the fallback chain (no image → initial bubble → native glyph restore) is solid. Fixes:
1. **Suppress session-row avatars when only one profile exists.** The JS decorates *every* row (`_sessProf[sid] || _activeName`), but the value ("which agent owns this chat") only exists with ≥2 profiles. In the common single-profile case you get an identical bubble stamped on every row — ~18px of title width, zero information, on a crown-jewel surface. One-line guard. (Also align the CSS comment, which says "other-profile rows," with actual behavior.)
2. Drop `margin-right:2px` on `.pa-session-avatar` — the row's `gap:6px` already spaces it (currently 8px after the avatar vs 6px elsewhere).
3. Replace `alert()` upload/delete errors with inline text in the modal (matches the app's polish).
4. Raise initial-bubble contrast — white 9px on `hsl(h,55%,45%)` falls to ~2.2:1 for yellow-greens; cap lightness ~38% or luminance-switch the text color.
5. Move focus into the manager dialog on open.

**Crown-jewel screenshots for sign-off:** composer footer (full + collapsed `cf-icons`, avatar set — confirm the +2px doesn't flip collapse); conversations list (single-profile-no-image case + multi-profile with a pinned row); transcript badges (image + native-glyph fallback, small & xlarge font sizes); titlebar chip (light+dark).

I'd like to see the sidecar auth reworked (shared credential) + the XSS/consent/size fixes, then I'll re-gate. Happy to help with the proxy-credential pattern if useful — it's the same fix all three need.

nesquena-hermes:
## Review — `sysinfo` (3 of 3)

@asorourx — the injection-hardening here is genuinely careful (allowlisted actions, list-form subprocess, timeouts, TTL caches — all verified). But this is the highest-privilege extension of the three (it can stop/restart/recreate the user's containers), and both gates found real issues, so verdict: **kick back for rework.** Codex live-verified an unauthenticated config change against the sidecar, and Fable found a functional feedback bug. Details below, then the **shared cross-cutting issue** that ties all three extensions together.

### 🔴 Security — must fix (Codex)
1. **No caller auth on host-mutating routes (`sysinfo.py:294-328,351`).** The sidecar runs speed tests + Docker mutations without authenticating the caller. Codex verified a **direct unauthenticated `Content-Type: text/plain` POST to port 17796 changed speed-test config.** For an extension that can stop/restart/recreate containers, that's the one that matters most. See the shared fix below.
2. **Docker mutations don't enforce the inventory allowlist (`docker_stats.py:449-485,617-686`).** `docker_action` accepts any 1–256-char docker-name-like value (not the "hex-ish ID" the comment claims — it'd accept leading-option-looking values), and `docker_group_action` accepts any regex-matching project. `docker_update` can inspect + recreate **any** compose-managed service, even one hidden from the UI. (You can't inject a *new* arbitrary image, but you can pull/recreate any existing targeted service.) Fix: resolve every id/project against the *filtered* inventory, require `[0-9a-fA-F]{12,64}` IDs, reject leading-option values, constrain compose workdirs to an approved root.
3. **`onclick` label breakout XSS (`sysinfo.js:405-419`).** Compose project labels go into an inline `onclick` string via an HTML escaper — but `&#39;` decodes back to `'` before the handler runs, so a crafted Docker label breaks out of the string. Remove inline handlers; use delegated listeners with opaque `data-*` values, and validate labels before returning inventory.
4. **DoS surface (`sysinfo.py:107-120,249-292`; `docker_stats.py:600`).** Unbounded request bodies, no single-flight guard on speed tests, unlimited rename keys, and bulk-update state not reserved under the lock before the worker starts. Cap bodies, bound rename entries to real inventory, add operation semaphores, set `running:true` atomically.

### 🟠 Breakage
5. **Proxy 10s timeout vs long operations (`sysinfo.js` + `docker_stats.py:390-511`).** Core proxies sidecars with a hard **10s** upstream timeout (`api/routes.py:5615`), but speed tests take 15–120s, registry checks repeat 25s calls, and compose updates allow 600s. The browser gets a 502 while the sidecar keeps mutating the host — inviting retries + concurrent mutations. Convert these to bounded single-flight background jobs that return immediately + expose a polling endpoint.
6. **Update state keyed by ID vs name mismatch (`docker_stats.py:495-503`, `sysinfo.js:732`).** Results/cache-invalidation use the container ID while update records + the frontend map are keyed by name, so a successful update leaves stale "update available" state. Resolve the name, call `updates_forget(name)`, return it.
7. **README inaccuracies:** state-file list omits `.docker_updates.json` (`docker_stats.py:297`); README claims "double-click to rename" but there's no `dblclick` handler.

### 🎨 UX (Fable — SHIP-WITH-UX-FIXES; one of these is really a functional bug)
1. **`showToast()` signature bug — failure feedback is invisible (18 call sites, `sysinfo.js:498-740`).** Core's signature is `showToast(msg, ms, type)`, but the extension calls `showToast(msg, 'error')` / `showToast(msg, 'info')`, so the type string lands in the **duration** slot → `setTimeout(fn,'error')` → NaN → 0ms → the toast vanishes in one frame. Every "Docker stop failed" / "Update failed" / bulk-update progress toast **never displays.** Fix: `showToast(msg, 4000, 'error')` (or omit both — core auto-classifies).
2. **Hand-rolled auto-schedule modal breaks light theme (`sysinfo.js:191-223`)** — hard-coded dark hex (`#151823`…) instead of core CSS vars / `appDialogOverlay`; renders as a dark alien box in light theme, no Escape/focus-trap. Restyle with vars or port to the core dialog. (Regresses the PR #58 light-theme requirement.)
3. **Three referenced keyframes don't exist** (`mc-st-pulse`, `mcDockerPulseOk`, `mcDockerPulseWarn` in `sysinfo.css`) — the "green pulse = running" dot silently doesn't animate. Define them behind `prefers-reduced-motion`, or drop the `animation` props.
4. **Background poll never stops (`sysinfo.js:768`)** — `setInterval(_siPollDocker, 20000)` shells `docker stats` every 20s even when the tab is hidden / Insights isn't showing. Core's own health poll gates on `document.visibilityState` + `main.showing-insights` — mirror that.
5. **A11y:** the "N updates" pill is a `<span onclick>` nested inside the collapse `<button>` (invalid interactive nesting, keyboard-unreachable, triggers the most destructive bulk action); group headers are `role=button tabindex=0` with no keydown handler; stack-header ▶↻■ buttons are ~21px (<24px WCAG 2.5.8). Make the pill a real `<button>` outside the toggle, add Enter/Space, bump targets to ≥24px.

**Screenshots for sign-off:** Insights tab with both cards expanded in dark + light theme (light exposes UX fix #2), plus one ≤700px mobile Docker-card shot. (Also: `extension.json` ships `"screenshots": []` and the dir is empty — the gallery will have no preview.)

---

### 🔴🔴 SHARED cross-cutting issue — all three extensions (the root of two KICK-BACKs)
This is the big one and it's the same in all three, so fixing it once (as a pattern) unblocks the set:

**(a) Sidecars disable their own auth while assuming the proxy is the only reachable path.** `rss-feeds/sidecar/shim.py:45-50`, `profile-avatars/sidecar/shim.py:121-132`, and sysinfo's handler all skip auth on the theory that only the WebUI proxy can reach them. But ports 17796/17797/17798 stay **directly reachable on the host** — Codex verified live unauthenticated writes against them. Anything else on the box (another local process, another extension's sidecar, a localhost-scoped SSRF) gets unauthenticated feed mutation / avatar writes / **container control**. The core proxy deliberately strips cookies/auth before forwarding, so the stubbed WebUI-cookie checks can't work — this needs a **real proxy-injected per-install secret** that the sidecar validates on every non-`/health` request. That's the fix for the two kick-backs.

**(b) All three auto-grant their own sidecar-proxy consent** (`sysinfo.js:25-30,750`, `feeds-inject.js:40-45,138`, `avatars.js:23-29,473`) by posting `approved:true` at init — silently bypassing the explicit user approval the install metadata promises. Remove it everywhere; leave each extension inactive with a clear "approve in Settings → Extensions" state until the user consents.

---

### Summary across the PR
| Extension | Security gate | UX gate |
|---|---|---|
| `rss-feeds` | Ship-with-fixes (SSRF + `esc()` XSS + redirect key-leak + disclosures) | Ship-with-UX-fixes (5 small) |
| `profile-avatars` | **Kick back** (auth hard-off + auto-consent + stored XSS + 512KiB cap) | Ship-with-UX-fixes (single-profile guard) |
| `sysinfo` | **Kick back** (no caller auth + allowlist bypass + `onclick` XSS + DoS) | Ship-with-UX-fixes (invisible-toast bug + light-theme modal) |

The bones are good and the sidecar reference pattern is valuable — but the shared proxy-auth boundary + the auto-consent removal are prerequisites for all three, and the two kick-backs need their per-extension security items too. Address the shared fix + per-extension lists and I'll re-run both gates on the new head and confirm they flip to clean. Happy to pair on the proxy-credential pattern since it's the common unblock. Thanks again for pushing the sidecar pattern forward — it's genuinely useful groundwork.

asorourx:
Thanks @nesquena-hermes 


 this is one of the most useful reviews I've had. Both gates (Codex security + Fable UX), full repo context, and live-verifying the unauthenticated loopback writes is exactly the scrutiny this pattern needs. I'd much rather find these now than ship them.
 
I accept the verdict: rss-feeds ship-with-fixes, profile-avatars and sys info kicked back — and I've turned the whole review into a tracked checklist.

On the shared proxy-credential boundary: I'll take you up on the pairing offer, since it's the real unblock for both kick-backs and it's core-side. 

Before I wire the sidecar validation I want to match your intended proxy side:
How is the per-install secret provisioned / where does it live (env the WebUI passes to the sidecar at start? a file? proxy-minted per install?)

What does the proxy inject on each forwarded request (header name?), and is /health exempt by design so the diagnostics probe still works?
Sidecars reject 401 on missing/mismatch — and is there a rotation/versioning story?

Once that's settled I'll enforce it on every non-/health request across all three sidecars.

Starting now on the parts that don't depend on it:
Removing the auto-granted proxy consent everywhere (each extension goes inactive with an explicit "approve in Settings → Extensions" state) — agreed, that shouldn't have been automatic.

rss-feeds security: SSRF is global+re-pin on fetch paths; DDN-API/txtContent rewrite to kill the esc() XSS + http(s)-only link constraint; redirect credential-leak fix (origin-lock + Gemini key + header).

The disclosure-honesty pass. On the rss remote-ollama SSH tunnel specifically — you're right it's surprising in an RSS reader; I'm inclined to remove that path from the gallery build rather than declare/harden native-host, unless you'd prefer it kept and hardened.

Two logistics questions: (1) keep this as one PR, or split into three so the ship-with-fixes rss-feeds can land independently of the two kick-backs? (2) I'll push fixes to this same branch and re-request review when the head's ready then gates to re-run — happy to go incremental if you prefer.


Thanks for pushing the sidecar pattern forward with me rather than bouncing it.

nesquena-hermes:
Great questions, @asorourx — and thank you for taking it in the constructive spirit it was meant. Let me answer each concretely. I went and read the actual core proxy source (`api/extensions.py` + `api/routes.py` at current `master`) so these are grounded in what core does *today*, not what I'd assume.

## The honest headline on the proxy-credential boundary

**There is no proxy→sidecar credential mechanism in core today.** I traced the whole forward path (`_extension_sidecar_proxy_request_headers`, `routes.py:5486`): core **strips** `authorization`, `cookie`, `host`, `origin`, `referer`, and `x-csrf*` before forwarding, and **injects nothing** in their place. It also does **not** start your sidecars — they're user-run manually per your install docs — so there's currently **no channel** (env or otherwise) through which core hands a secret to a sidecar at start.

So the boundary you'd validate against doesn't exist yet — it's a **core change**, and because it's a security-architecture decision that affects every future sidecar extension, it's a **maintainer call on the core side** (I don't want to hand you a header contract I invented and have it change under you). Let me get that design locked with the maintainer before you wire the sidecar side. In the meantime, here's my concrete recommendation so you can see the shape and we're designing to the same target:

**Recommended design (subject to Nathan's sign-off):**
- **Provisioning:** core mints a per-install random token, persists it in the extensions state file (alongside `sidecar_proxy_consents`), and exposes it to the sidecar **at the moment of the manual install/run** — most likely as an env var the install step sets (e.g. `HERMES_EXT_SIDECAR_TOKEN`) written into the printed run command / the `.service` file the README already ships. (Core-managed sidecar lifecycle would be cleaner but is a much bigger change; the env-at-manual-run path fits today's model.)
- **Injection:** the proxy adds a header on each forwarded request — proposed `X-Hermes-Sidecar-Token: <token>` (final name is Nathan's to bless). Since core controls both ends, it's a shared secret, not a bearer token exposed to the browser.
- **`/health` exemption:** **yes, keep `/health` unauthenticated by design** — the diagnostics probe hits it directly cross-origin, so it must stay open (and, per my rss-feeds comment, it also needs `Access-Control-Allow-Origin: *`). Everything except `/health` requires the token.
- **Sidecars reject 401 on missing/mismatch** — correct.
- **Rotation/versioning:** keep it simple v1 — a single current token, regenerated on demand (a "reset sidecar token" action) which just invalidates old sidecar processes until they're restarted with the new env. No multi-key ring needed for a single-user tool; I'd punt versioning unless Nathan wants it.

**Bottom line:** don't build the sidecar-side validation until we lock the header name + provisioning channel — I'll confirm within the next cycle. Everything else below is independent and you can start now (which matches your plan).

## Your two specific sub-answers
- **512 KiB is real and load-bearing.** `_EXTENSION_SIDECAR_PROXY_MAX_RESPONSE_BYTES = 512 * 1024` (`routes.py:283`). That's why the profile-avatars 1 MiB cap breaks — cap both ends at 512 KiB.
- **The 10s upstream timeout is real too.** `opener.open(request, timeout=10)` (`routes.py:5615`) — hard 10s on every proxied call. That's the root of the sysinfo speedtest/compose-update 502s; the background-job-with-polling refactor is the right fix, not a bigger timeout (you can't change core's 10s from the extension).

## The remote-ollama SSH tunnel (rss-feeds)
Your instinct is right — **remove it from the gallery build.** An RSS reader spawning `ssh -N -L` with `accept-new` TOFU is a scope surprise that'll cost you a `native_host:true` declaration + a hardening review for a path most users won't use. Dropping it keeps rss-feeds a clean `loopback-sidecar` reference and lets it stay `native_host:false` honestly. If you want the remote-model path later, it's a much better fit as its own opt-in extension where the SSH scope is the *point* and can be declared up front.

## Your two logistics questions
1. **Split into three PRs — yes, please.** rss-feeds is ship-with-fixes and can land independently once its (non-shared) security items are done; the two kick-backs shouldn't hold it hostage. Practically: keep this PR as rss-feeds, open two new ones for profile-avatars and sysinfo. Note the registry `README.md` conflict — each entry touches it, so whichever lands first will make the others show CONFLICTING; just rebase the trailing ones (I can help sequence that).
   - **Caveat:** even split out, rss-feeds still needs the shared proxy-token boundary before it's truly shippable (its sidecar is directly reachable too). So the realistic order is: land the core-side token mechanism first, then rss-feeds rides it. But splitting now still helps — you can get rss-feeds fully green *except* the token, and it merges the moment core lands.
2. **Same branch + incremental is perfect.** Push fixes to each branch, re-request review, and I'll re-run **both gates** (Codex security + Fable UX) on the new head each round and confirm they flip to clean. Incremental is genuinely better here — the shared boundary is the kind of thing worth converging in a round or two rather than one big drop.

## Suggested order of attack (matches what you're already doing)
1. **Now, no dependencies:** remove auto-consent everywhere (inactive + "approve in Settings" state); rss-feeds SSRF (`is_global` + connect-to-validated-IP + re-pin every hop on both `_http_get` and `_fetch_article_text`); the `esc()`→DOM/`textContent` + `http(s)`-only link rewrite; the OpenRouter/Gemini redirect credential fix; drop the SSH path; all the disclosure/README + `extension.json` honesty fixes; the sysinfo `showToast` + light-theme-modal + keyframes + visibility-gated poll UX fixes; the profile-avatars single-profile guard + stored-XSS `textContent` fix + 512 KiB cap.
2. **After the token contract is blessed:** wire the `X-Hermes-Sidecar-Token` (or final name) validation on every non-`/health` route across all three sidecars — the single change that clears both kick-backs.

I'll come back with the confirmed header name + provisioning channel shortly. Ping me any time — happy to pair on the sidecar validation once the contract's locked. Thanks again for pushing this pattern forward.

asorourx:
That makes perfect sense Nathan.  Looking back, I was trying to validate a trust boundary that doesn't actually exist in the current architecture.

I'll hold off on implementing the sidecar validation until the proxy-side contract (provisioning + injected header) is defined in core. In the meantime I'll keep working through the independent fixes so everything is ready once that boundary exists.

Thanks for digging through the core implementation and confirming the current behavior—it saved me from building against an interface that isn't there yet.

asorourx:
@nesquena-hermes Split done as you suggested — this PR is now **rss-feeds only**, with
profile-avatars (#66) and sysinfo (#67) as their own PRs.

The independent rss-feeds fixes are all pushed to this branch:
- auto-consent removed (reads `/api/extensions/status`, shows an "approve in
 Settings" state)
- SSRF hardening — reject any non-global IP, then pin the connection to the
validated IP and re-pin on every redirect hop, on both feed fetch and article
retrieval
- XSS — quote-safe `esc()`, http(s)-only link rewriting, delegated listeners
- credential-leak on redirect — redirects disabled on the credentialed LLM
calls; Gemini key moved to the `x-goog-api-key` header
- dropped the SSH tunnel path (stays a clean loopback sidecar, `native_host:false`)
- disclosure — README enumerates every sidecar egress + filesystem read;
`network_external:true`; all `mc.feeds.*` storage keys declared
- `/health` CORS — `ACAO:*` + `no-store`, scoped to `/health` only

Only remaining item is the shared proxy→sidecar token, which is on your side to
define — happy to wire it the moment the header/provisioning contract lands.
Ready for another gate run on the new head whenever you are.

asorourx:
All the independent review items are addressed and pushed to this branch:
- auto-consent removed (reads `/api/extensions/status`, shows an "approve in Settings" state)
- SSRF hardening — reject any non-global IP, then pin to the validated IP and re-pin on every redirect hop, on both feed fetch and article retrieval
- XSS — quote-safe `esc()`, http(s)-only link rewriting, delegated listeners
- credential-leak on redirect — redirects disabled on the credentialed LLM calls; Gemini key moved to the `x-goog-api-key` header
- dropped the SSH tunnel path (stays a clean loopback sidecar, `native_host:false`)
- disclosure — README enumerates every sidecar egress + filesystem read; `network_external:true`; all `mc.feeds.*` storage keys declared
- `/health` CORS — `ACAO:*` + `no-store`, scoped to `/health` only

Only the shared proxy→sidecar token is left, which is core-side (yours to define) — happy to wire it the moment the header/provisioning contract lands. Ready for another gate run.

nesquena-hermes:

@asorourx, the proxy-to-sidecar authentication boundary we were waiting on is now implemented and merged in both repositories:

- Core enforcement: [nesquena/hermes-webui#6331](https://github.com/nesquena/hermes-webui/pull/6331)
- Authoritative contract: [`docs/SIDECAR_CONTRACT.md`](https://github.com/hermes-webui/hermes-webui-extensions/blob/main/docs/SIDECAR_CONTRACT.md)
- Drop-in Python scaffold: [`examples/sidecar-scaffold/`](https://github.com/hermes-webui/hermes-webui-extensions/tree/main/examples/sidecar-scaffold)

Please rebase this branch onto current `main` first. PR #65 added the metadata rules and CI checks this entry now needs to satisfy.

### Wiring `rss-feeds` to the standard

1. **Declare a vendored token-v1 runtime in `extension.json`:**

   ```json
   "sidecar": {
     "type": "loopback",
     "origin": "http://127.0.0.1:17797",
     "health_path": "/health",
     "proxy_auth": "token-v1",
   }
   ```

   Repeat `"proxy_auth": "token-v1"` in the generated/runtime `manifest.json` sidecar block, but omit the library-only `runtime` object there.

2. **Adopt the canonical scaffold instead of the current custom server.** Copy `sidecar_base.py` and `sidecar.py` byte-for-byte from `examples/sidecar-scaffold/`. Add:

   ```json
   { "id": "rss-feeds", "port": 17797, "proxy_auth": "token-v1" }
   ```

   as `sidecar/sidecar.json`. Put the extension-owned route registrations in `sidecar/routes_impl.py`. The current `ThreadingHTTPServer` in `sidecar/sidecar.py` cannot remain beside the scaffold; the usage gate rejects any second listener because it would bypass the deny-by-default token check.

3. **Remove the fake auth shim.** The current `shim.is_auth_enabled() -> False` posture should disappear. Auth is now inherited at the scaffold's single dispatch point. Every non-`/health` request requires the core-injected `X-Hermes-Sidecar-Token`; missing/wrong token returns 401, and an absent token file returns 503. The sidecar never mints or embeds the token.

4. **Replace the refresh SSE route.** `/api/feeds/refresh/stream` cannot survive through the sidecar proxy because the proxy buffers responses and has an approximately 10-second timeout. Convert it to start-job + poll, for example `POST /api/feeds/refresh` returning a job id immediately plus a short `GET /api/feeds/refresh/{job_id}` status/result route. The existing background summary pattern is a good model.

5. **Update the systemd unit.** It should run the byte-compared entrypoint from the declared runtime directory:

   ```ini
   WorkingDirectory=%h/.hermes/webui/extensions/rss-feeds/sidecar
   ```

   Do not use `/usr/bin/env`, do not put the token in the unit, and keep extension-specific state configuration separate if still needed.

6. **Keep the client on the fixed same-origin proxy path and keep consent manual.** The auto-consent removal you already pushed is correct. `token-v1` also fails closed with 403 while WebUI authentication is off, so the inactive state should tell the operator to enable **Settings → Password**, then grant sidecar consent under **Settings → Extensions**.

Core #6331 first appears in `exp-v0.52.129`; until that reaches the stable channel, the README should say “requires a WebUI build containing #6331” rather than guessing a stable minimum.

Before re-requesting review, please run the normal entry gates plus:

```bash
```

Once this lands, the shared auth blocker from the earlier review is resolved structurally rather than through per-route checks. The prior RSS-specific security and UX items still need to remain fixed on the rebased head.

asorourx:
`token-v1` adopted per [`docs/SIDECAR_CONTRACT.md`](../blob/main/docs/SIDECAR_CONTRACT.md) — rebased onto current `main`.

- vendored `sidecar_base.py` + `sidecar.py` **byte-identical**; `sidecar.json` = `{id, port: 17797, proxy_auth: "token-v1"}`
- `routes_impl.py` maps the routes onto the existing hardened `feeds.py` via a small capture adapter, so the ~1900-line reader logic (SSRF/redirect hardening, credentialed LLM calls) stays byte-for-byte intact behind the scaffold's token guard
- dropped the SSE `/api/feeds/refresh/stream` route — the proxy buffers, and the frontend already uses the plain `POST /api/feeds/refresh` (its EventSource path was already dead-coded)
- `extension.json` + runtime `manifest.json` declare `proxy_auth: token-v1` (runtime: vendored); `.service` uses `/usr/bin/python3 -S -u sidecar.py`, no token in the unit
- README: token-v1 + same-UID scope + docker limitation

Both gates pass locally. Functionally verified (temp port + token): `/health` tokenless; deny without/with-wrong token (401); list/settings/entries + gzip work; a bad feed URL still 400s through the adapter.

nesquena-hermes:
## Review — `rss-feeds` (maintainer QA pass)

Thanks for this, @asorourx — it's an ambitious extension and the token-v1 sidecar wiring is right. I ran the full gate on the frozen head (`afc3fd3`): static + sandboxed Codex, Fable UX, and a sandboxed runtime harness. The reader front-end is high-craft, but there are **merge-blocking security + reliability issues in the sidecar** that need rework before this can land. Grouped by severity, each verified against source:

### 🔴 Security
1. **SSRF bypass on favicon fetch — `sidecar/feeds.py:1493`.** Favicon retrieval uses redirect-following `urllib.request.urlopen()` instead of your own IP-validating, connection-pinning `_safe_fetch()`. A provider redirect can reach a non-global address, which contradicts the README's "every fetch is SSRF-checked" guarantee. Route **both** favicon requests through `_safe_fetch()` so every hop is revalidated + pinned.
2. **Unbounded response reads (memory DoS) — `sidecar/feeds.py:1130` + backends at `165,204,256,329`.** Redirect bodies and all model-backend responses are read with unbounded `resp.read()`. A hostile/large peer can exhaust memory, and oversized output later blows the proxy envelope. Use one shared capped reader for every backend/redirect response.
3. **Unbounded thread spawning — `sidecar/feeds.py:1877`.** Every summarize request spawns an unrestricted daemon thread, and each concurrent refresh spawns 8 workers. Add a bounded queue/semaphore + single-flight refresh; reject or coalesce once the bound is hit.
4. **Stored-link XSS on legacy rows — `assets/feeds.js:2222`.** New ingestion rejects unsafe schemes, but stored summary source URLs reach `href` via `_attr()` **without** `_safeUrl()`, so a pre-existing `javascript:`/`data:` link in an older DB stays clickable. Use `_attr(_safeUrl(s.link))` and sanitize existing stored links on read.

### 🟠 Breakage
5. **Consent contract mismatch — `assets/feeds-inject.js:46`.** Current core returns sidecars under **top-level** `status.sidecars`, but this reads `status.extensions[*].sidecars` and then treats the missing record as consented. The approval screen never appears, and with WebUI auth off users get failing proxy calls instead of the "enable Settings → Password, then approve" guidance. Match `status.sidecars` by `id`, require `proxy.consented === true`, fail closed on a missing record, and surface `proxy.posture === "local_unprotected"`.
6. **`feedparser` unimportable under the mandated `python3 -S` — `sidecar/rss-feeds-sidecar.service:13`.** The canonical service command uses `-S` (no site-packages), but `feeds.py:1182` imports the separately-`pip install`ed `feedparser`, so add/refresh **just fails** on a clean install. Keep `-S` and either vendor a reviewed parser under `sidecar/` or parse with stdlib; update README install steps to match.
7. **Synchronous long POST vs the ~10s proxy timeout — `sidecar/feeds.py:1659`.** Refresh is a synchronous 10–30s+ POST; `summary-test` (`:1678`) can block up to ~240s. This was the exact start-job/poll change requested in the earlier review round but not preserved. Move refresh / summary-test / slow add-validation to bounded job-start + poll routes; don't hold the proxy request open.
8. **Response can exceed the 512 KiB proxy cap — `sidecar/feeds.py:1401`.** The API allows 2,000 entries × ~6 KiB attacker-controlled title/summary; the front-end asks for 200–300. High-entropy content can exceed the encoded-response ceiling. Add server-side pagination + a hard byte budget below 512 KiB; bound `done_entry_ids` and per-item model output.

### 🟡 Disclosure (permission/README honesty — the install model trusts this)
9. **`extension.json:43` understates API authority.** `webui_api.read` is empty despite reading `/api/extensions/status` and all proxied GETs; `write` lists only an *unused* consent endpoint while the extension POST/PATCH/DELETEs sidecar state. Enumerate the real read/write routes + `/api/extensions/status`; drop `/api/extensions/sidecar-proxy-consent` unless it's actually called.
10. **`extension.json:66` — `filesystem.arbitrary:false` is false.** `feeds.py:117-140,215-235` reads the non-owned credential file `~/.hermes/.env`. Either set `filesystem.arbitrary:true` and disclose it, or drop the file fallback and take keys only from the process environment.
11. **README corrections.** (a) `≥ 0.16` is wrong — this needs a build containing core #6331 (`exp-v0.52.129`+); (b) the "feed GETs carry no query params" claim contradicts `feeds.py:1124-1125`, which forwards the subscribed URL's query; (c) "summaries never use paid models" isn't enforceable for Gemini (keys inherit the project's billing tier). Fix all three.

### Verdict
**Bounce for rework** — the SSRF/DoS/XSS trio (#1–#4) plus the `feedparser`-under-`-S` break (#6) and the synchronous-POST timeout (#7) are individually merge-blocking. The front-end reader itself is strong; once the sidecar is hardened and the job/poll refactor lands, I'm happy to re-gate. Take your time — no rush. Happy to clarify any item.

*(Gate: Codex static+sandboxed, Fable UX, sandboxed runtime. No PR code executed outside the no-network/no-secrets sandbox.)*

nesquena-hermes:
## Exact-head gate provenance correction + UX addendum

A delayed background-process notification exposed a class-level issue in the original review run: the first Codex command used a temporary worktree that the extensions reviewer wrapper rejects. The older generic RSS Codex/Fable logs were from the pre-split combined branch, not recoverable proof for this exact PR head. The source reproductions and bounce verdict were valid, but the parenthetical claim that frozen split head `afc3fd3` had completed both reviewer legs was too strong.

I have now rerun **both reviewers from scratch** in a full extensions-repo clone at the still-current head `afc3fd3c95930de561f16c403bfdc6be5eaa0133`. Because this extension contains outbound networking and credential paths, both reruns were strictly static/read-only; no PR source was executed.

### Exact-head results

- **Codex: KICK BACK TO AUTHOR.** It independently reproduced every one of the eleven findings in my previous comment: the unguarded favicon fetch, unbounded response reads/deadlines, unbounded worker creation, stored legacy-link execution, wrong consent-status shape, `feedparser` incompatibility with the declared `python3 -S` runtime, synchronous proxy-incompatible network operations, response-size ceiling, and permission/filesystem/README mismatches. It found no reason to retract or weaken that feedback.
- **Fable: PASS on placement/design, SHIP-WITH-UX-FIXES overall.** It confirmed the reader is correctly housed as a self-owned overlay, keeps the composer/chat/sidebar/workspace crown jewels untouched, adapts well to mobile, and now uses theme tokens and honest indeterminate progress. These are strengths worth preserving through the sidecar rework.

Fable also found current-head UX follow-ups that I manually verified and that are not duplicates of the security/reliability list:

1. **Visible icon-only controls have no visual tooltips — `assets/feeds-inject.js:76-89`.** Search, filters, all-feeds, clicked, summaries, settings, add, refresh, and close have `aria-label`s but do not use the host's `has-tooltip`/`data-tooltip` convention. Several icons are novel enough that sighted users cannot infer them reliably. Apply the same custom-tooltip system already used by the launcher.
2. **Overlay modal/focus/Escape behavior is incomplete — `assets/feeds-inject.js:58-137`.** The card lacks `aria-modal`, initial focus, and focus restoration. Its Escape handler closes the top layer correctly but does not stop propagation, so host Escape handlers can silently close Settings/search/dropdowns underneath. Consume the event when Feeds handles it and restore focus to the launcher on close.
3. **Desktop drag-region + launcher target — `assets/feeds.css:21-35`, `assets/feeds-inject.js:141-146`.** The launcher opts out of titlebar dragging, but `.hx-feeds-card` does not; under window-controls-overlay its top strip can become a native drag surface. Add `-webkit-app-region: no-drag`. Also target the real `.app-titlebar` directly rather than the nonexistent `.app-titlebar-right` primary selector.
4. **Reduced motion is incomplete — `assets/feeds.css:354-380,619-623,1035-1043,1275-1281`.** The progress bar is covered, but the title marquee, `mcSumPulse`, and `mc-ring-pulse` remain infinite under `prefers-reduced-motion`. Disable all three there.
5. **Successful refresh feedback is computed then discarded — `assets/feeds.js:1994-2000,2020-2027`.** `finish()` passes the useful `N/M feeds · K new` message to `_feedProgress.done(msg)`, but `done()` accepts and renders nothing. Surface that final result once rather than silently dropping it. In the auth-off approval state, also name the complete remedy: enable **Settings → Password**, then approve under **Settings → Extensions**.

The disposition remains **bounce for rework**, now backed by valid exact-head Codex + Fable evidence. The hard blockers remain the sidecar's SSRF/resource/job/runtime boundaries; the UI architecture itself is sound, and the items above can be folded into the same pass without redesigning the reader.

nesquena-hermes:
## Independent-review addendum — one additional current-head finding

A separate static reviewer completed after the exact-head Codex/Fable pass. I reconciled its report against the existing bounce and manually verified one non-duplicate item at unchanged head `afc3fd3c95930de561f16c403bfdc6be5eaa0133`:

### Browser and sidecar independently schedule the same automatic refresh

- `sidecar/feeds.py:606-609,720-799` starts a backend daemon from the first settings read and calls `refresh_all()` when its interval expires.
- `assets/feeds.js:427-480` independently runs a one-second browser timer and invokes the same refresh POST when its cached timestamp expires.
- Closing the Feeds overlay only hides it (`assets/feeds-inject.js:133-137`); it does not stop the browser timer.
- The backend thread-creation lock protects only daemon creation, not `refresh_all()` itself.

Once a user has opened the overlay, the hidden browser and backend daemon can reach the same deadline and run overlapping outbound fetches/SQLite writes. This compounds the already-posted synchronous-proxy and unbounded-worker issues.

**Fix shape:** make the sidecar the sole automatic-schedule authority and let the browser display/poll its state; keep the browser path for explicit manual refresh only. Independently enforce one shared refresh single-flight/job coordinator server-side so auto and manual requests cannot overlap. If any browser scheduling remains, stop it when the overlay is torn down/hidden.

This folds into the existing start-job/single-flight rework; it does not change the bounce disposition.

asorourx:
Thanks for the exact-head gate — all findings are addressed on the new head (`b0de95b`). Mapping to your list:

**🔴 Security**
1. **SSRF favicon** — `_fetch_favicon` now goes through `_safe_fetch` (IP-validated, connection-pinned, per-hop revalidated), same as every other fetch.
2. **Unbounded reads** — one shared `_read_capped` (2 MiB) on every model backend + the ollama tags read; the redirect drain is bounded to 64 KiB.
3. **Unbounded threads** — summarize is capped by a `BoundedSemaphore` (rejects with 429 past the bound); `refresh_all` is single-flight; the auto-fetch daemon skips rather than blocks.
4. **Stored-link XSS** — `_renderSources` runs every stored link through `_safeUrl` before `href`, sanitizing legacy `javascript:`/`data:` rows on read.

**🟠 Breakage**
5. **Consent contract** — reads top-level `status.sidecars`, matches by id, requires `proxy.consented === true`, **fails closed** on a missing record, and surfaces `posture: local_unprotected` (→ enable Password, then approve).
6. **feedparser under `-S`** — replaced with a stdlib `xml.etree` parser (RSS 2.0 / Atom / RSS 1.0); no pip dep, verified under `python3 -S`.
7. **Sync long POSTs** — refresh, summary-test, and add-validation are all start-job + poll now; nothing holds the proxy request open.
8. **512 KiB cap** — entries response is byte-budgeted (450 KiB, `truncated` + `next_before` cursor, `before` param for older pages); stored summary output capped at 32k and `done_entry_ids` bounded.

**🟡 Disclosure**
9. `webui_api` now enumerates the real proxied read/write routes + `/api/extensions/status`; dropped the unused consent endpoint.
10. Dropped the `~/.hermes/.env` reads entirely — keys come only from the process env, so `filesystem.arbitrary:false` is honest.
11. README: requires a build with #6331 (`exp-v0.52.129`+); corrected the query-params claim; dropped the unenforceable "no paid models" (added the Gemini-billing caveat).

**Addenda** — the sidecar daemon is the sole auto-refresh scheduler (browser observes/polls only, timer torn down on hide) and `refresh_all` is single-flight, so auto+manual can't overlap. Fable items folded in: custom tooltips on the icon controls, `aria-modal` + focus in/out + capture-phase Escape that no longer leaks to host handlers, `-webkit-app-region:no-drag` + the real `.app-titlebar` target, reduced-motion on the marquee/pulse/ring/spinner, and the refresh result toast is surfaced instead of dropped.

All extension + sidecar-contract gates pass locally. Ready for re-gate whenever you have a cycle — thanks again.

franksong2702:
## Current-head follow-up — background job ownership

Thanks for the hardening work at `b0de95b`. I confirmed the token-v1 scaffold, SSRF re-validation plus IP pinning, redirect-safe credentialed model calls, bounded fetch reads, and the add-feed background-validation direction are present.

One merge blocker remains: refresh jobs have no identity. The sidecar keeps one global `_refresh_job`; if refresh A finishes and refresh B starts before A's UI polls its terminal state, B overwrites A's result. The frontend then waits on and renders B's status/results as if they belonged to A. This can show the user the wrong operation outcome.

Please return a `job_id` when starting a refresh, make the status endpoint query that exact job, and retain bounded terminal results for a short TTL. Add a regression for: A fails -> B starts -> A's poll still receives A's failure.

The summary-model Test flow has the same global-job shape. Please give it the same ownership treatment, so concurrent tabs/configurations cannot display another test's result.

Small follow-ups for the same update:
- The README says the sidecar reads `~/.hermes/.env`, while the implementation intentionally reads credentials only from its process environment. Make the filesystem disclosure match the code.
- Add the requested user-visible screenshots; `extension.json` currently has an empty `screenshots` array.

Once these land, I will re-run the delta gate.

asorourx:
Confirmed, and thanks for the exact repro. The `_refresh_job` singleton is the same ownership gap wearing a different hat — B's refresh overwrites A's result and the UI renders B's status against A's poll, so the user can watch the wrong refresh's outcome. And you're right that the summary-model **Test** flow is the same global-job shape, so it needs the identical treatment or the bug just relocates.

What I'll push:
- Refresh returns a `job_id`; the status endpoint queries that exact job; terminal results held in a small bounded/TTL map so a slow poller still gets its own outcome. The frontend waits on the id it was handed, nothing else.
- Same per-invocation ownership applied to the summary-model Test flow — no shared global — so concurrent tabs/configs can't cross-render.
- Regression: A fails → B starts → A's poll still receives A's failure.

Follow-ups: the **`.env` disclosure** is a doc bug — the sidecar reads creds from its process environment only, so I'll fix the README to match (env-only, no `~/.hermes/.env` read); and I'll populate the empty `screenshots` array (reader two-pane light + dark, mobile full-screen overlay with the thumb cluster).

Re-requesting once the head's green for the delta gate. Good catch on the job-identity gap before it shipped.

asorourx:
@franksong2702 — pushed the fixes; current head is `d457c2b`.

- **Merge blocker (job ownership):** refresh and the summary-model Test each carry an id now. `start_refresh` / `start_summary_test` return `{id}`, `refresh_status(id)` / `summary_test_status(id)` return *that* job's own record from a bounded/TTL buffer keyed by id, and the frontend polls `?id=` and waits only on its own job — B finishing before A's poll can no longer overwrite A's result or render B's status as A's. Regression `test_job_ownership.py` covers both flows (A fails → B starts → A's poll still receives A's failure).
- **Follow-up:** README filesystem disclosure corrected — the sidecar reads Summarize creds from its own process environment only, not `~/.hermes/.env` (matches `feeds.py`).

Still open: the **screenshots** (reader two-pane light + dark, and the mobile full-screen overlay). Those need the live authenticated UI, so I'll push them separately. Code-side is all on this head — re-run the delta gate whenever works for you.

asorourx:
Update: screenshots landed — head is now `d7933f4`, with the two-pane reader (light + dark) in `screenshots/` and wired into `extension.json`. Full branch is ready for the delta gate.

franksong2702 (approved):
Reviewed the current head d7933f4. Refresh and summary-test jobs now preserve id-scoped terminal results, with regression coverage for the prior cross-job result mix-up. Repository checks are green.

PR #65, opened 2026-07-19
feat: sidecar scaffold + contract (companion to core token-v1 auth)

nesquena-hermes:
## What this does

Adds the canonical sidecar scaffold and repository contract for `loopback-sidecar` extensions. It is the library-side companion to the proxy-to-sidecar `token-v1` boundary in **nesquena/hermes-webui#6331**.

Core owns consent, per-extension token provisioning, proxy target validation, and `X-Hermes-Sidecar-Token` injection. This PR owns the language-neutral metadata contract, the optional Python reference scaffold, repository validation, tests, examples, and CI enforcement.

## Runtime ownership model

Sidecar discovery is manifest-driven and distinguishes two valid cases:

1. **Repository-vendored runtimes**
   - Declare `sidecar.runtime.kind: "vendored"` and a safe relative runtime path.
   - Use `proxy_auth: "token-v1"`.
   - Vendor byte-identical canonical `sidecar_base.py` and `sidecar.py` files.
   - Keep extension routes in a regular `routes_impl.py` file with exactly one valid `register(app)` hook.
   - Launch through exactly one validated `/usr/bin/python3` command with `-S`,
     optional `-u`, and `sidecar.py` from the declared/default extension entry.

2. **External runtimes**
   - Declare `sidecar.runtime.kind: "external"` and the runtime repository.
   - May use any language, framework, or layout.
   - Implement the language-neutral token/header/health/rotation contract when using `token-v1`.
   - Remain explicitly reported as `legacy` until migrated.

Every runtime kind uses core-compatible endpoint syntax: a bare loopback origin
and a normalized absolute health path, both without a query or fragment.

Desktop Companion is modeled as the existing external compatibility case. Its Node/Tauri runtime remains external and explicitly legacy; this PR does not force it into the Python scaffold.

## Canonical scaffold

`examples/sidecar-scaffold/sidecar_base.py` owns dispatch and validates authentication at one chokepoint. Only `GET/HEAD /health` is tokenless. It returns direct-browser-readable CORS and `Cache-Control: no-store` headers. Every other route fails closed for missing, wrong, stale, or unavailable token material. Tokens are read per request so rotation takes effect without restarting the sidecar.

The scaffold supports path parameters, query strings, raw and multipart bodies, JSON and binary responses, custom headers, gzip JSON, and an optional background hook. `sidecar.py` remains a thin canonical entrypoint, `sidecar.json` holds per-entry identity/port/auth settings, and `routes_impl.py` contains extension-owned behavior.

## Enforcement

The shared contract library now drives synchronization, validation, usage checks, and adversarial tests:

- Metadata defines the target set. Omitting the protected scaffold cannot make a vendored sidecar disappear.
- Unknown authentication modes fail closed, and `extension.json`, runtime `manifest.json`, and `sidecar.json` must agree.
- Protected scaffold/config/route paths must be real regular files.
- Symlinks anywhere beneath an extension root are rejected before their targets
  are read, including declared assets and every component or nested entry of
  `runtime.path`, so validation and artifact collection inspect the same tree.
- Symlinked extension roots, directories with missing or dangling metadata, and
  non-object JSON metadata are surfaced as structured validation failures rather
  than being silently omitted or raising an unstructured exception.
- Python package, extension-module, and bytecode names that could shadow protected modules are rejected.
- Vendored services must use one unprefixed command with the pinned interpreter
  and `-S`. The working directory must match the default state-tree installation
  or an exact extension entry declared with `HERMES_EXT_INSTALL_DIR`. Custom
  paths must be absolute or `%h`-anchored and emit an explicit reviewer warning.
  WebUI's ambiguous `HERMES_WEBUI_EXTENSION_DIR` is deliberately not reused.
  PATH wrappers, arbitrary Python-looking
  executables, direct shebang execution, Python command/module flags, trailing
  commands, duplicate launch commands, non-foreground service types, auxiliary
  `Exec*`, `.service.d` drop-ins, unsafe execution-context directives, and
  unverifiable Quadlet entrypoints are rejected.
- The route hook is checked with Python AST. Bare decorator-factory calls,
  unreachable or conditional registration, malformed signatures, decorators on
  `register`, duplicate/direct/wildcard/pattern rebindings, rebinding through
  evaluated definition expressions or module-level comprehensions, statically
  visible `global register` declarations, module-level control flow, and
  unconditional module raises fail. The docs state the static check's unavoidable
  limit for dynamic namespace mutation.
- Server source is scanned across every extension entry, including malformed or incomplete declarations, for unguarded listener construction in the supported source languages.
- Canonical server-file exemptions require byte identity in the same usage check.

## Tests and CI

The workflow runs:

- entry validation and validator self-tests;
- adversarial sidecar-contract tests;
- real-socket Python scaffold tests covering health CORS/cache headers, missing/wrong/rotated/deleted tokens, optional background behavior, and unknown auth modes;
- extension safety scanning;
- canonical scaffold synchronization checks;
- sidecar listener/service usage checks;
- Desktop Companion and loopback-example agreement checks;
- registry generation and JSON validation.

Generated ZIP artifacts set the standard UTF-8 filename flag in both local and
central headers, with a cross-runtime regression test for non-ASCII asset names.

Focused fixtures also pin origin/health-path parity with core, including
rejection of numeric, hexadecimal, and percent-encoded loopback spellings that
JavaScript's URL parser would otherwise normalize into acceptance.

The protected scaffold directory, loopback example, and contract documentation
are included in CI path filters. Registry generation independently reruns the
sidecar contract checks before packaging any artifact.

## Verification

The complete workflow-equivalent sequence passes locally on exact commit `a0fb6175b44dc61a439ecffe70d2573b7209c71e` across all 14 extension entries, including registry artifact generation and parsing, Desktop Companion validation, Node syntax checks, and `git diff --check`.

## Sequencing

Core PR **nesquena/hermes-webui#6331** supplies the token-v1 proxy enforcement. This PR defines and enforces the matching library contract. Neither PR is merged by this update.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.github/workflows/extensions.yml
+++ b/.github/workflows/extensions.yml
@@ -6,6 +6,7 @@ on:
       - 'extensions/**'
       - 'scripts/**'
       - 'examples/sidecar-scaffold/**'
+      - 'examples/loopback-sidecar/**'
       - 'README.md'
       - 'docs/extension-entry.md'
       - 'docs/SIDECAR_CONTRACT.md'
@@ -18,6 +19,7 @@ on:
       - 'extensions/**'
       - 'scripts/**'
       - 'examples/sidecar-scaffold/**'
+      - 'examples/loopback-sidecar/**'
       - 'README.md'
       - 'docs/extension-entry.md'
       - 'docs/SIDECAR_CONTRACT.md'
@@ -69,6 +71,9 @@ jobs:
 
       - name: Check sidecar usage (no rogue servers / correct ExecStart)
         run: node scripts/check-sidecar-usage.mjs
+
+      - name: Validate Desktop Companion and loopback example
+        run: node scripts/validate-desktop-companion.mjs
 
       - name: Generate registry
         run: node scripts/generate-registry.mjs --out dist/registry.json
--- a/.github/workflows/extensions.yml
+++ b/.github/workflows/extensions.yml
@@ -5,8 +5,10 @@ on:
     paths:
       - 'extensions/**'
       - 'scripts/**'
+      - 'examples/sidecar-scaffold/**'
       - 'README.md'
       - 'docs/extension-entry.md'
+      - 'docs/SIDECAR_CONTRACT.md'
       - 'docs/rfcs/0001-extension-json-schema.md'
       - '.github/workflows/extensions.yml'
   push:
@@ -15,8 +17,10 @@ on:
     paths:
       - 'extensions/**'
       - 'scripts/**'
+      - 'examples/sidecar-scaffold/**'
       - 'README.md'
       - 'docs/extension-entry.md'
+      - 'docs/SIDECAR_CONTRACT.md'
       - 'docs/rfcs/0001-extension-json-schema.md'
       - '.github/workflows/extensions.yml'
 
@@ -40,11 +44,22 @@ jobs:
         with:
           node-version: '22'
 
+      - name: Set up Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: '3.12'
+
       - name: Validate extension entries
         run: node scripts/validate-extensions.mjs
 
       - name: Test extension validator
         run: node scripts/test-extension-validator.mjs
+
+      - name: Test sidecar contract enforcement
+        run: node scripts/test-sidecar-contract.mjs
+
+      - name: Test canonical sidecar scaffold
+        run: python scripts/test-sidecar-scaffold.py
 
       - name: Scan extension safety gates
         run: node scripts/scan-extension-safety.mjs
--- a/.github/workflows/extensions.yml
+++ b/.github/workflows/extensions.yml
@@ -49,6 +49,12 @@ jobs:
       - name: Scan extension safety gates
         run: node scripts/scan-extension-safety.mjs
 
+      - name: Check sidecar scaffold is in sync
+        run: node scripts/sync-sidecar-base.mjs --check
+
+      - name: Check sidecar usage (no rogue servers / correct ExecStart)
+        run: node scripts/check-sidecar-usage.mjs
+
       - name: Generate registry
         run: node scripts/generate-registry.mjs --out dist/registry.json
 
--- a/README.md
+++ b/README.md
@@ -90,8 +90,8 @@ Extension entries should document:
 - manual verification steps
 - any future CI check that should protect the extension
 
-The long-term goal is for this repository to validate existing extension
-entries as the main WebUI extension contract evolves.
+This repository validates existing extension entries as the main WebUI extension
+contract evolves.
 
 Run the current repo-wide checks locally with:
 
@@ -122,6 +122,7 @@ The one-click **install flow and registry UI shipped in core** (Settings →
 Extensions: browse the registry, review permissions, install/uninstall with
 sha256 verification). Merged entries here are validated + safety-scanned in CI,
 published to the registry, and become one-click-installable from inside WebUI.
-The Tier-A loopback sidecar proxy remains future/demand-driven; sidecar-class
-entries currently reach their local backend directly under the loopback CSP
-allowance and should document their manual install + lifecycle explicitly.
+The consent-gated per-extension loopback proxy is shipped. `token-v1` runtimes use
+that same-origin path and validate core's per-install token; explicitly legacy
+external adapters may still use direct loopback under the core CSP allowance.
+Every sidecar entry must document its install and lifecycle requirements.
--- a/README.md
+++ b/README.md
@@ -98,11 +98,16 @@ Run the current repo-wide checks locally with:
 ```bash
 node scripts/validate-extensions.mjs
 node scripts/test-extension-validator.mjs
+node scripts/test-sidecar-contract.mjs
+python3 scripts/test-sidecar-scaffold.py
 node scripts/scan-extension-safety.mjs
+node scripts/sync-sidecar-base.mjs --check
+node scripts/check-sidecar-usage.mjs
+node scripts/validate-desktop-companion.mjs
 node scripts/generate-registry.mjs --out dist/registry.json
 ```
 
-Pull requests run the same validator and safety scan in CI. Pushes to `main`
+Pull requests run the same validation, contract, and safety checks in CI. Pushes to `main`
 generate the registry and per-extension zip artifacts for GitHub Pages. The
 registry entry includes a `download` URL and artifact-level `sha256` for each
 extension so the WebUI install client fetches and verifies reviewed bytes before
--- a/VISION.md
+++ b/VISION.md
@@ -50,19 +50,15 @@ Extensions grow in capability in stages. Each rung is independently shippable.
 2. **Settings UI** ✅ *shipped* — a Settings → Extensions surface: listing,
    enable/disable, and one-click install/uninstall from the gallery (with sha256
    verification against the registry), plus a diagnostics tab.
-3. **Sidecar metadata + direct loopback** *foundation merged (descriptive metadata); direct loopback shipped* — an extension can
-   *declare* a local helper process (`sidecar: { type, origin, health_path }`) so
-   WebUI can show the dependency and, later, report coarse health. **Today**, an
-   extension's JS can talk to a trusted loopback sidecar **directly** — the core
-   CSP `connect-src` already allows `http://127.0.0.1:*` / `localhost` / `ws://`
-   loopback — so direct browser → sidecar is the *current* sidecar-class pattern.
-4. **Backend routes / same-origin proxy** *planned / future* — a **future, opt-in,
-   per-extension same-origin proxy** (`/extensions/<id>/sidecar/*` → the declared
-   loopback origin). This is *not* required for sidecar extensions to work today
-   (see #3 — direct loopback already works); it's future core work for cases the
-   direct path can't cover (e.g. the sidecar's own CORS, hiding the port, uniform
-   same-origin auth). Repo entries should **not** depend on it yet. In-process
-   server route handlers are a later, separately-gated step beyond that.
+3. **Sidecar metadata + direct loopback** ✅ *shipped* — an extension can declare
+   a local helper process (`sidecar: { type, origin, health_path }`) for health and
+   dependency diagnostics. An explicitly legacy external adapter may still use
+   direct browser → loopback access under core's loopback CSP allowance.
+4. **Same-origin sidecar proxy** ✅ *shipped* — core provides an opt-in,
+   consent-gated per-extension proxy (`/api/extensions/<id>/sidecar/*` → the
+   declared loopback origin). `token-v1` runtimes use this path and authenticate
+   every non-health request with the per-install token injected by core. In-process
+   server route handlers remain a later, separately-gated step beyond the proxy.
 
 ## The registry & install experience (target)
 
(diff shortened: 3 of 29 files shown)


nesquena-hermes:
Companion core PR (the token-v1 mechanism this scaffold validates against): nesquena/hermes-webui#6331

@franksong2702 @rodboev — could you both take a look and provide any feedback when you get a chance? Frank, this touches the extension/sidecar surface you know best — especially interested in your read on the scaffold ergonomics (does `routes_impl.py` + `sidecar.json` feel right for authors?) and the CI usage-lint approach. The full contract is in `docs/SIDECAR_CONTRACT.md`.

franksong2702 (changes requested):
Thanks Nathan. I reviewed head `81e762311b0693c9d7ed958f869b1b379a41946e`, focusing on the two areas you called out: scaffold ergonomics and whether the CI usage lint actually enforces the boundary.

The `routes_impl.py` + `sidecar.json` split is a reasonable authoring model: extension code stays small, and authentication belongs at one shared dispatch point. I am requesting changes because three current behaviors break the security and runtime guarantees this PR documents.

### 1. Unknown `proxy_auth` values fail open

`examples/sidecar-scaffold/sidecar_base.py:142` accepts any string, and lines 183-185 authorize every request whenever the value is not exactly `token-v1`. In an isolated checkout I changed the config to `"proxy_auth": "token-vI"` and invoked a registered route without a token; the route returned `200 {"secret": true}`.

Please validate the enum at startup and fail closed on every unknown value. Since this is the canonical scaffold for new sidecars, CI should also verify that `sidecar.json` and the extension manifest agree on the auth mode.

### 2. The usage lint cannot discover a sidecar that omits the scaffold

`scripts/check-sidecar-usage.mjs:41-45` and `scripts/sync-sidecar-base.mjs:20-27` define a sidecar extension as a directory that already contains `sidecar_base.py`. That makes the file being enforced also the discovery signal an implementation can omit.

I added an isolated `extensions/desktop-companion/sidecar/rogue.py` fixture containing `ThreadingHTTPServer(...)`, without `sidecar_base.py`. The full validator, safety scan, sync check, and usage lint all exited 0; the new checks reported `0 sidecar extension(s)`.

Please derive the target set from the extension manifest/capability source of truth, require the canonical files for applicable sidecars, and keep an adversarial missing-scaffold fixture in the lint self-tests. The server scan should not depend on the canonical file already being present.

### 3. The scaffold health response violates the documented browser contract

`docs/SIDECAR_CONTRACT.md:67-70` requires `/health` to return `Access-Control-Allow-Origin: *` and `Cache-Control: no-store`, because WebUI probes the sidecar with a direct cross-origin browser fetch. `sidecar_base.py:196-198` currently returns only `Content-Type: application/json`. The endpoint returns HTTP 200 to curl, but the WebUI browser probe cannot read it and will report blocked/unreachable.

Please add both headers and cover the actual cross-origin health response in a scaffold test.

There are also two contract wording mismatches worth cleaning up in the same PR: the docs name `auth_required` while core #6331 returns `posture`, and the token section says eager provisioning while the current core implementation provisions on consent/resolve.

Validation performed on the isolated PR checkout:

- scaffold Python files compile successfully;
- `node scripts/validate-extensions.mjs` passes all 14 entries;
- `node scripts/scan-extension-safety.mjs` passes with the rogue fixture present;
- both new sidecar scripts exit 0 and incorrectly report zero sidecars with that fixture;
- typo-auth dispatch returns 200 without a token;
- `/health` dispatch returns only `{"Content-Type": "application/json"}`.

Once these three blockers are fixed, I am happy to re-review the narrowed diff.

franksong2702:
## Compatibility clarification: external sidecar runtimes

I want to clarify the manifest-driven discovery recommendation in my review so the fix does not accidentally turn into a Python-only requirement for every `loopback-sidecar` declaration.

Desktop Companion is the concrete compatibility case:

- `extensions/desktop-companion/extension.json` declares the `loopback-sidecar` capability and `127.0.0.1:17787`, but this repository does not vendor its server implementation;
- the entry points to `franksong2702/hermes-webui-desktop-companion`, where the actual sidecar is a Node HTTP server and the source-built Tauri app also talks to that runtime directly;
- its manifest currently omits `proxy_auth`, so core treats it as legacy. A future token-v1 migration would require moving the WebUI adapter behind the core proxy and adding equivalent token validation to the Node dispatch path. Copying the Python scaffold is not a viable drop-in migration for that product.

So the line in the PR body saying that the extensions library has zero shipped sidecars is true only in the narrow sense that this repository does not ship the runtime source. The library already ships an external sidecar declaration, and the contract/validator needs to model that distinction explicitly.

Suggested enforcement model:

1. **Repository-vendored token-v1 sidecars:** discover them from manifest metadata, require the canonical scaffold and entrypoint, and byte-compare the protected files.
2. **External sidecar runtimes:** require explicit external-runtime ownership metadata (exact field name is open for design) and a language-neutral token-v1 conformance contract. Python can remain the canonical reference implementation without becoming the only permitted runtime language.
3. **No discovery escape:** server code committed anywhere under an extension entry must still be scanned. Omitting `sidecar_base.py` must not make an in-repository rogue server invisible.
4. **Legacy remains explicit:** an external runtime that has not completed token-v1 migration should remain clearly reported as legacy rather than being silently treated as scaffold-compliant.

This does not weaken the missing-scaffold blocker from the review. It separates two valid cases that CI currently conflates: an intentionally external runtime versus in-repository server code that omitted the guard. Please capture this distinction in `docs/SIDECAR_CONTRACT.md` and in the validator tests before tightening discovery.

nesquena-hermes:
Implemented the full follow-up for Frank's review and compatibility clarification. The frozen successor commit is `a0fb6175b44dc61a439ecffe70d2573b7209c71e`.

## Frank review mapping

### 1. Unknown `proxy_auth` values failed open

Resolved in the canonical scaffold and validators:

- `examples/sidecar-scaffold/sidecar_base.py` validates the auth enum at startup and rejects unknown values.
- `scripts/extension-registry-lib.mjs` and `scripts/sidecar-contract-lib.mjs` require agreement among `extension.json`, runtime `manifest.json`, and `sidecar.json`.
- `scripts/test-sidecar-scaffold.py`, `scripts/test-extension-validator.mjs`, and `scripts/test-sidecar-contract.mjs` cover typo/unknown modes and declaration drift.

### 2. Missing scaffold escaped discovery

Resolved by making manifest metadata authoritative:

- `sidecar.runtime.kind` explicitly distinguishes `vendored` from `external`.
- `scripts/sidecar-contract-lib.mjs` discovers declared sidecars from metadata, then requires the protected files for vendored runtimes.
- Entry validation and registry generation both run the scaffold and usage
  contract checks, so an empty declared vendored runtime cannot be packaged.
- The source scan still walks every extension entry, including incomplete or malformed declarations, so omitting `sidecar_base.py` cannot hide a committed server.
- Adversarial fixtures cover missing scaffold files, hidden runtime declarations, and rogue source outside the expected runtime directory.

### 3. `/health` violated the browser contract

Resolved and tested over a real socket:

- `GET/HEAD /health` now returns `Access-Control-Allow-Origin: *` and `Cache-Control: no-store`.
- `scripts/test-sidecar-scaffold.py` verifies those response headers through HTTP, along with 401/503 behavior, valid-token access, per-request token rotation, and token deletion.

### 4. Contract wording drift

Resolved in `docs/SIDECAR_CONTRACT.md` and the examples:

- Uses the core `posture` vocabulary.
- Describes provisioning on consent plus lazy reprovision during proxy resolution.
- Separates current proxy support from browser-direct integrations.
- Documents the proxy timeout/body envelope, same-UID trust scope, auth-off posture, and bridge-network Docker limitation without overstating isolation.

## External-runtime compatibility clarification

Implemented as requested:

- Repository-vendored token-v1 runtimes use the canonical Python scaffold.
- External runtimes declare their repository and may use any language or runtime layout while satisfying the language-neutral wire contract.
- Desktop Companion is explicitly `external` and `legacy`, with matching origin, health, and auth metadata. It is not forced into the Python scaffold.
- Unknown runtime/auth values fail closed, while an external legacy runtime remains visible as a warning instead of being silently treated as compliant.

The PR body has also been rewritten to remove the stale “zero shipped sidecars” and “Node-only CI” claims and to describe this ownership split accurately.

## Additional convergence hardening

The repeated independent reviews found and drove tests for several adjacent bypass classes:

- no unconditional `start_background(app)` call and no unguarded second listener;
- broad defensive listener scanning, including Node aliases and common Python, Go, Rust, Java, Kotlin, C#, and Ruby forms;
- section-aware systemd parsing and a small `[Service]` directive allowlist;
- exact `/usr/bin/python3 -S [-u] sidecar.py` launch validation, with PATH wrappers, arbitrary interpreters, direct shebang execution, Python command/module flags, trailing tokens, auxiliary `Exec*`, execution-root overrides, and vendored Quadlet units rejected;
- real-file and byte-identity checks for protected files;
- scaffold synchronization uses link-aware checks before writing, so a dangling
  protected-file symlink cannot redirect canonical bytes outside the runtime;
- fail-fast symlink rejection for every entry beneath an extension root,
  including declared assets and nested runtime entries that would be omitted
  from generated artifacts, before validation reads an external target;
- explicit discovery errors for symlinked extension roots and missing/dangling
  `extension.json`, plus a structured error for non-object JSON metadata;
- ZIP local and central headers mark member names as UTF-8, with a Python
  `zipfile` round-trip regression for non-ASCII assets;
- rejection of package, compiled-extension, and bytecode names that could shadow canonical Python modules;
- AST validation for exactly one undecorated top-level synchronous
  `register(app)` binding with an actual direct route decorator application,
  rejecting bare `app.route(...)`, false/conditional paths, malformed
  signatures, duplicate/imported/assigned/wildcard/pattern rebindings, rebinding
  while evaluating later definitions or module-level comprehensions, and
  statically visible `global register` declarations, module-level control flow,
  and unconditional module raises. The docs also state the static check's limit
  for dynamic namespace mutation;
- exact install-directory/working-directory agreement. Non-default installs use
  the sidecar-only `HERMES_EXT_INSTALL_DIR` for the exact extension entry rather
  than overloading WebUI's ambiguous `HERMES_WEBUI_EXTENSION_DIR`;
- custom state/install directories must be absolute or `%h`-anchored and produce
  an explicit reviewer warning, plus rejection of ignore-failure prefixes,
  non-foreground service types, duplicate launch commands, and `.service.d`
  drop-ins;
- core-compatible endpoint validation for every runtime kind, including rejection
  of path/query-bearing values and alternate numeric, hexadecimal, or encoded
  loopback spellings that JavaScript would otherwise normalize;
- the standalone vendored-runtime contract applies the same raw canonical-origin
  rule rather than relying on URL-parser normalization;
- explicit validator output requiring reviewers to confirm custom roots are
  operator-controlled rather than broadly writable;
- CI path filters include the loopback example, and CI directly verifies that
  Desktop Companion and the external/legacy example stay in agreement.
- The metadata RFC now records the implemented two-file model with exact drift
  checks, and stale pre-gallery installation wording has been removed.

## Verification on the frozen commit

All passed:

- `node scripts/validate-extensions.mjs` — 14 entries
- `node scripts/test-sidecar-contract.mjs`
- `python3 scripts/test-sidecar-scaffold.py -v` — 3 real-socket tests
- `node scripts/sync-sidecar-base.mjs --check`
- `node scripts/check-sidecar-usage.mjs`
- `node scripts/generate-registry.mjs --out dist/registry.json` plus JSON parse
- `node scripts/validate-desktop-companion.mjs`
- Node syntax checks for every touched script
- `git diff --check`

Core enforcement remains in **nesquena/hermes-webui#6331**; this PR does not weaken or duplicate core consent, target validation, token provisioning, or proxy authentication.

@franksong2702 @rodboev, this changed materially in response to review, so a final re-review is requested before merge. I have not merged either PR.

franksong2702 (approved):
Re-reviewed current head `a0fb6175b44dc61a439ecffe70d2573b7209c71e`.

The three blockers from my previous review are resolved:

- unknown `proxy_auth` values now fail closed in the scaffold and validator coverage;
- sidecar discovery is manifest-driven, with explicit `vendored` vs `external` runtime ownership, and committed rogue server code no longer escapes by omitting `sidecar_base.py`;
- `/health` now returns the browser-readable CORS/cache headers and is covered by real-socket tests.

The Desktop Companion external legacy model is also captured correctly: it remains visible as an external runtime without forcing the Node/Tauri implementation into the Python scaffold.

Validation I ran:

- `node scripts/test-sidecar-contract.mjs`
- `python3 scripts/test-sidecar-scaffold.py -v`
- `node scripts/sync-sidecar-base.mjs --check`
- `node scripts/check-sidecar-usage.mjs`
- `node scripts/generate-registry.mjs --out dist/registry.json` plus JSON parse
- `node scripts/validate-desktop-companion.mjs`
- Node/Python syntax checks and `git diff --check`

I do not see remaining merge blockers. Approving.

PR #66, opened 2026-07-19
Add profile-avatars extension

asorourx:
## Profile Avatars

Gives each Hermes profile its own avatar image everywhere the profile shows up:
the profile chips, the assistant badge in the transcript, and the session list
(including sessions that belong to other profiles). Avatars are stored
server-side by a small loopback sidecar, so a picture set on one device appears
on every device with **zero `localStorage`**.

### How it works
- A `manifest-bundle` frontend (`assets/avatars.js` + CSS) that decorates the
  existing profile/session UI — no core views are replaced.
- A `loopback-sidecar` on `127.0.0.1` that stores and serves the images,
  reachable only through the WebUI's consented sidecar proxy (same-origin to the
  browser; the sidecar never faces the network).
- Uploaded images are downscaled and size-bounded before storage.

### Items being addressed on this branch
- single-active-profile guard (only decorate the profile(s) actually in use)
- render profile names/handles via `textContent` (no `innerHTML`) to close a
  stored-XSS path
- 512 KiB image cap on both ends, matching the core sidecar-proxy limit
- declare the full read surface (sessions, including archived) in `extension.json`
- UX polish pass

### Dependency: shared proxy→sidecar token
The core sidecar proxy currently forwards requests without any proxy→sidecar
credential, so there is no trust boundary for a sidecar to validate against yet.
Provisioning + an injected header is a **core-side** change being defined by the
maintainer. Sidecar-side token validation will be wired here the moment that
contract lands; every item above is independent of it and handled on this branch.


asorourx:
Pushed the independent kickback fixes on this branch:
- stored-XSS closed — the manager list builds name/sub with `textContent`
instead of `innerHTML` (the label can be the user-set bot name)
- 512 KiB cap on both client and server `_MAX_BYTES` to match the core proxy
response cap (verified: a 600 KB upload now 413s up front instead of storing
then 502-ing on read); README's 1-MiB references fixed
- `extension.json` now declares the real API surface: `/api/extensions/status`
plus the proxied avatar list/blob reads and POST/DELETE writes; dropped the
unused `sidecar-proxy-consent` write
- single-profile guard — session-row avatars only render with ≥2 profiles
- archived mis-attribution fixed — one note here: I went to wire the "supported
archived pagination params" but core's `/api/sessions` actually honours only
`?all_profiles=1` (limit/offset/include_archived are all ignored — I checked
`routes.py`). So rather than send params that do nothing, I stopped
attributing unknown session rows to the active profile: archived/other rows
the payload doesn't include now get no bubble instead of the wrong one. If
there's a different endpoint that paginates archived sessions, point me at it
and I'll populate them properly.

Only the shared proxy→sidecar token piece is left, which waits on the core
contract. Ready for a re-run on the new head.

asorourx:
All the kickback items are addressed and pushed:
- stored-XSS closed — manager list builds name/sub via `textContent`, not `innerHTML`
- 512 KiB cap on both client and server (matches the core proxy response cap; a 600 KB upload now 413s up front instead of storing then 502-ing on read); README's 1-MiB refs fixed
- `extension.json` declares the real API surface (`/api/extensions/status` + the proxied avatar list/blob reads and POST/DELETE writes); dropped the unused `sidecar-proxy-consent` write
- single-profile guard — session-row avatars only render with ≥2 profiles
- archived mis-attribution fixed — core `/api/sessions` honours only `?all_profiles=1` (limit/offset/archived are ignored — I checked `routes.py`), so unknown rows now get no bubble instead of the active profile's; if there's a different endpoint that paginates archived sessions, point me at it and I'll populate them.

Only the shared proxy→sidecar token remains (core-side). Ready for another gate run.

nesquena-hermes:

@asorourx, the core and library pieces for the proxy-to-sidecar credential are now merged:

- Contract: [`docs/SIDECAR_CONTRACT.md`](https://github.com/hermes-webui/hermes-webui-extensions/blob/main/docs/SIDECAR_CONTRACT.md)
- Scaffold: [`examples/sidecar-scaffold/`](https://github.com/hermes-webui/hermes-webui-extensions/tree/main/examples/sidecar-scaffold)

Please rebase onto current `main` first so this branch includes the new validators.

### Wiring `profile-avatars`

1. Add the new fields to `extension.json`:

   ```json
   "sidecar": {
     "type": "loopback",
     "origin": "http://127.0.0.1:17798",
     "health_path": "/health",
     "proxy_auth": "token-v1",
   }
   ```

   Repeat `proxy_auth: token-v1` in the runtime `manifest.json` sidecar block; `runtime` stays only in `extension.json`.

2. Replace the custom `sidecar/sidecar.py` with the byte-identical canonical `sidecar.py`, add the byte-identical `sidecar_base.py`, and create:

   ```json
   { "id": "profile-avatars", "port": 17798, "proxy_auth": "token-v1" }
   ```

   as `sidecar/sidecar.json`.

3. Move the avatar routes into `sidecar/routes_impl.py`:

   - `GET /api/avatars`
   - `GET /api/avatars/{profile}`
   - `POST /api/avatars/{profile}`
   - `DELETE /api/avatars/{profile}`

   The scaffold provides `req.params`, `req.headers`, and raw `req.body`, and handlers return `app.json(...)` or `(status, headers, bytes)`. The multipart parser therefore needs to consume `req.body` rather than `handler.rfile`. Keep the 512 KiB avatar cap and the current MIME/ETag/nosniff checks.

4. Delete the current auth-off shim path (`is_auth_enabled() -> False`, cookie stubs, and per-route `_require_auth`). The scaffold checks `X-Hermes-Sidecar-Token` once at dispatch, so list/read/write/delete are all protected without a route author remembering a guard. `/health` is the only tokenless route and is owned by the scaffold.

5. Update `profile-avatars-sidecar.service` to the required execution shape:

   ```ini
   WorkingDirectory=%h/.hermes/webui/extensions/profile-avatars/sidecar
   ```

   No `/usr/bin/env`, no token in the unit. Core provisions `STATE_DIR/sidecar-auth/profile-avatars.token` when consent is granted; the scaffold resolves and re-reads it per request, so rotation is live.

6. Keep the browser on `/api/extensions/profile-avatars/sidecar/...` and keep consent manual. The auto-consent removal you already pushed is right. With WebUI auth off, token-v1 consent/proxy access returns 403 by design; tell the operator to enable **Settings → Password** and then approve the sidecar under **Settings → Extensions**.

Core #6331 first appears in `exp-v0.52.129`. Until stable promotion, document the requirement as “a WebUI build containing #6331.”

Please run the normal entry gates plus:

```bash
```

A useful local proof for this entry is: `/health` succeeds without a token; avatar list/blob/upload/delete fail directly with no/wrong token; all succeed through the authenticated WebUI proxy; deleting the token file flips protected routes to 503; restoring/rotating it takes effect without restarting.

asorourx:
`token-v1` adopted per [`docs/SIDECAR_CONTRACT.md`](../blob/main/docs/SIDECAR_CONTRACT.md) — rebased onto current `main` (the #65 scaffold).

- vendored `sidecar_base.py` + `sidecar.py` **byte-identical**; `sidecar.json` = `{id, port: 17798, proxy_auth: "token-v1"}`
- routes moved into `routes_impl.py` behind the scaffold's deny-by-default guard; `/health` is the only tokenless route
- `avatars.py` reduced to pure storage + a bytes-based multipart parse; dropped the old `shim.py` and all route-local auth
- `extension.json` + runtime `manifest.json` both declare `proxy_auth: token-v1` (runtime: vendored)
- `.service` runs `/usr/bin/python3 -S -u sidecar.py`, no token in the unit, `HERMES_WEBUI_STATE_DIR` only
- README documents token-v1, the honest same-UID scope, and the docker loopback limitation

Both gates pass locally: `sync-sidecar-base.mjs --check` + `check-sidecar-usage.mjs`. Functionally verified on a temp port with a token file: `/health` tokenless 200; missing/wrong token → 503/401; list/upload/fetch/delete all work through the guard.

asorourx:
Fixed the `Validate extensions` failure: the rebased `main`'s validator maps `/api/sessions` usage to the short token `sessions`, so I declared `sessions` in `permissions.webui_api.read` (`5a3ca13`). All six `extensions.yml` steps now pass locally (`validate-extensions`, `test-extension-validator`, `test-sidecar-contract`, `scan-extension-safety`, `sync-sidecar-base --check`, `check-sidecar-usage`). The run is sitting in `action_required` — needs a workflow approval on your side to go green.

nesquena-hermes:
Merged in `2e784fe` — thanks @asorourx! 🎉 profile-avatars is the first vendored token-v1 sidecar to land in the gallery.

Full credit to you for the design and implementation. I applied maintainer hardening on the branch before merge (all gate-clean):
- **Consent contract** — read top-level `status.sidecars`, require `proxy.consented`, fail closed
- **Multipart parser** — swapped the byte-split for an `email.policy` parser (embedded-boundary binary payloads were truncated yet passed PNG sniffing)
- **Cache-buster** — fold the content ETag into `?v=` so same-second re-uploads don't strand a stale blob
- **Renamed-root aliasing** — normalize `default`/empty/renamed-root so a renamed root keeps its avatar on chips/badges/legacy rows
- **Archived sessions** — `include_archived=1&archived_limit=2000` so archived rows get their owner's avatar
- **Object-URL lifecycle** — revoke/evict superseded blob URLs
- **Modal a11y** — `aria-modal`, focus trap, capture-phase Escape, focus restore; `showToast` over `alert`; confirm-before-remove; `--surface`/`--error` theme tokens; 44px mobile tap targets
- **README** — corrected to require a build with core #6331 (`exp-v0.52.129`+)

Gate: Codex + Fable + sandboxed runtime harness + real browser matrix (desktop + mobile). I've left detailed rework notes on your other two (#64 rss-feeds, #67 sysinfo) — both solid concepts that need sidecar hardening before they can land.

nesquena-hermes:
Post-merge audit correction: a delayed background-process notification exposed that the first Codex invocation from this review had rejected its worktree. The valid replacement run had produced the repair set, but I could not preserve proof of a post-fix convergence pass on the final #66 head. I therefore reran Codex and Fable against the exact merged content rather than treating the notification as stale.

That rerun found real lifecycle and nested-dialog issues. They were fixed, tested, re-gated to convergence, and merged in follow-up #69 as `fe3e704`:
- late avatar-fetch ownership and object-URL cleanup;
- restoration after `custom-avatar` clears an existing badge;
- core confirmation stacking and Escape ownership in both listener orders;
- visible inline errors and stale-error lifecycle.

Final #69 evidence: repository CI green, isolated full gate green, real desktop/mobile Chromium green, Codex **SAFE TO SHIP**, Fable **SHIP-UX**. The public record should reflect that #66 was merged before this final convergence proof and #69 completed it. The original Profile Avatars contribution remains credited to @asorourx.

PR #69, opened 2026-07-21
Fix profile-avatar lifecycle and dialog ownership

nesquena-hermes:
## Summary

Corrective follow-up to #66, **Profile Avatars** by @asorourx. A delayed background-process notification exposed that the first Codex invocation in the original review had rejected its worktree. The valid replacement gate had driven the initial repairs, but a preserved post-fix convergence pass was missing before merge. Re-running both advisors against the exact merged head found real lifecycle and dialog-integration defects; this PR fixes them.

Only the extension's existing JS and CSS change.

## Fixes

- deduplicate in-flight avatar downloads by source URL;
- bind each completion to its requested URL so a late old fetch cannot overwrite a replacement/removal;
- revoke stale object URLs after upload/remove and after late completions;
- restore profile-avatar badges when `custom-avatar` clears an image on an existing role element;
- move the manager below the core dialog layer (`1000 < 1100`);
- use core's real `confirmLabel` option and preserve safe Cancel focus;
- yield Escape/Tab to a nested core dialog regardless of listener registration order (`defaultPrevented || dialogOpen`);
- render manager errors inline instead of behind the overlay as a toast;
- clear stale inline errors when the user starts a new upload/remove attempt.

## Verification

Exact reviewed head: `6df07f02feb7fd0e8f64e7c5d99bda613e626882`

- focused behavioral suite, including RED→GREEN/non-vacuity cases:
  - delayed old/new fetch ordering, in-flight dedupe, and stale URL revocation;
  - text-node-only `custom-avatar` removal mutation;
  - core dialog layering/API/keyboard contracts;
  - both key-listener registration orders;
  - inline-error lifecycle;
- full repository validators and safety checks in a no-network/no-secrets/no-`$HOME` sandbox: green;
- threat scan: clean, score 0;
- authenticated token-v1 proxy upload on current WebUI core: green;
- real Chromium desktop and 390×844 mobile:
  - core confirmation visibly above manager, correct **Remove** label, Cancel focused;
  - Escape closes only the nested dialog and leaves manager open;
  - forced sidecar error stays visible inline, no hidden toast;
  - clearing `custom-avatar` restores both existing transcript badges;
  - mobile manager/dialog fit; close/Replace/Remove are 44 px;
- screenshots independently vision-checked and sent to the maintainer;
- Codex final round: **SAFE TO SHIP**, no findings;
- Fable final round: **SHIP-UX**, no findings.

[...]

PR #70, opened 2026-07-21
Harden Profile Avatars upload and storage invariants

nesquena-hermes:
## Summary

Follow-up hardening for Profile Avatars after the late independent review of #66 and the lifecycle correction in #69.

- validate PNG/JPEG/WebP structure and dimensions server-side, with matching browser decode checks;
- enforce core-compatible profile IDs, a transaction-safe 128-row cap, symlink refusal, and owner-only SQLite modes;
- refresh the profile roster and recheck sidecar consent every time the manager opens;
- normalize image decode errors and raise generated-initial contrast to WCAG AA across all hues;
- accept legal quoted multipart boundaries;
- add a standard-library regression suite to repository CI.

Profile Avatars remains authored and credited to @asorourx (`extension.json` author is unchanged). This PR corrects hardening gaps discovered after #66 was merged; #69 remains the preceding lifecycle fix.

## Verification

- `python3 scripts/test-profile-avatars.py` (12 tests)
- complete local `.github/workflows/extensions.yml` command sequence
- `node --check extensions/profile-avatars/assets/avatars.js`
- previous #69 lifecycle/dialog harnesses plus new decode/manager runtime harness
- threat scan + no-network/no-secrets sandbox run
- token-v1 sidecar E2E: fail-closed auth, upload, byte-identical read, ETag/304, malformed/dimension/profile rejection, live token rotation, delete, DB mode
- Fable UX gate: one decode-message finding fixed with mutation-proven regression; final rerun pending
- Codex exact-head gate pending

### DIFF
--- a/.github/workflows/extensions.yml
+++ b/.github/workflows/extensions.yml
@@ -66,6 +66,9 @@ jobs:
       - name: Test canonical sidecar scaffold
         run: python scripts/test-sidecar-scaffold.py
 
+      - name: Test Profile Avatars
+        run: python scripts/test-profile-avatars.py
+
       - name: Scan extension safety gates
         run: node scripts/scan-extension-safety.mjs
 
--- a/extensions/profile-avatars/README.md
+++ b/extensions/profile-avatars/README.md
@@ -19,8 +19,8 @@ load** into a shared in-memory object URL. Nothing is written to
   to the title, so with "Show N from other profiles" you can see at a glance
   which agent each chat belongs to. Re-renders live when you switch profiles.
 - **Manager modal** — a button at the top of the **Profiles** tab opens an upload / replace / remove
-  panel for every profile. PNG / JPEG / WebP, up to 512 KiB, magic-byte sniffed
-  server-side.
+  panel for every profile. PNG / JPEG / WebP, up to 512 KiB and 4096px / 16
+  megapixels, decoded before upload and structurally validated server-side.
 - **Colored-initial fallback** — profiles without an image get a deterministic
   colored initial bubble.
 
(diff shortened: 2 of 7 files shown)


franksong2702:
## Current-head blocker — renamed root-profile identity

The upload/storage hardening is in good shape on `13146af`: token-v1, structural image validation, bounded storage, SQLite modes, symlink refusal, and the refreshed consent-aware manager are all present.

One merge blocker remains: avatar storage does not preserve Hermes Core's root-profile alias semantics. Core treats `default` and a renamed root profile (for example `kinni`, reported with `is_default: true`) as the same identity. The extension stores an existing avatar under `default`, but refresh looks up only `amap[p.name]`. After the root profile is renamed, it therefore looks for `amap["kinni"]`, fails to find the prior `default` record, and the avatar disappears. Re-uploading under the new display name leaves the old row behind and consumes another slot.

Please use one stable storage key for the root profile (the existing `default` alias is the natural choice), while rendering it under the current root display name. Apply that canonicalization consistently for lookup, upload/replace, and delete. Add a regression covering: avatar stored as `default` -> roster reports `{name: "kinni", is_default: true}` -> it remains visible and can be replaced/removed without creating an orphan row.

`rename_avatar()` exists but currently has no route or frontend caller, so it does not address this flow. Once this is fixed, I will re-run the delta gate.

nesquena-hermes:
## Maintainer fix pushed — exact-head gate passed on `773ea66ecde37cb4aea4202cb729b5f945d22846`

I rebased this PR onto current `main` and fixed the remaining renamed-root persistence class rather than only the public lookup symptom:

- both `default` and the renamed root display name resolve/render one entry;
- the durable sidecar key remains `default`, so an existing avatar survives a root rename;
- upload writes the stable key and cleans a legacy display-name row;
- remove deletes both stable and legacy aliases;
- empty/whitespace/unknown public API names are rejected before any sidecar request.

The VM harness now models a real backend row stored under `default`, proves the image survives the rename, verifies the exact POST/DELETE sequence and legacy cleanup, and covers invalid-name behavior. I also mutation-tested the oracle: replacing stable-root lookup with the old display-name lookup makes the new regression test fail with the expected “avatar stored under default did not survive root rename” error.

### Final gate

- GitHub Actions: **success** — https://github.com/hermes-webui/hermes-webui-extensions/actions/runs/30503928006
- Full local 16-entry workflow: **pass**
- Profile Avatars: **13/13 tests pass**
- Codex exact-head extension gate: **SAFE TO SHIP**, including its own mutation bite and route-sequence probe
- Fable exact-head UX gate: **SHIP-UX / no release-blocking UX issues**
- Worktree and generated-artifact cleanup: clean

This is a self-authored visible extension and the PR changes the workflow, so I am not merging it on agent-only evidence. @franksong2702 @rodboev, please do the final human review of the exact head, including the existing desktop/mobile visuals: composer fallback at wide+narrow widths, multi-profile session rows (including renamed-root legacy rows), transcript/custom-avatar coexistence, and the manager/error-banner states. If the existing media no longer proves those exact states, please request refreshed screenshots before approval.

franksong2702 (approved):
Independent exact-head review completed on `773ea66`: renamed-root alias read/write/delete paths are covered by the new regression suite, and extension validation, safety, sidecar contract, scaffold sync, and usage checks passed. Approved.

PR #75, opened 2026-08-02
feat(system-monitor): titlebar system monitor widget + audit conventions

ChonSong:
## Summary

Adds the **System Monitor** titlebar extension — an inline resource monitor (CPU / RAM / Disk / IO / Net) embedded directly in `.app-titlebar-inner`, with a braille-wave popup, two-color net split fill, adaptive net scale, compact K/M labels, and a settings panel. Includes a full audit pass for conventions and LOC reduction.

## What it does

- **Titlebar widget** — 5 square cells (CPU, RAM, Disk, IO, Net), fill-from-bottom, overlay number labels (no `%`), click to open popup.
- **Net split fill** — TX fills up from center, RX fills down, two distinct colors (`netUpColor` / `netDownColor`).
- **Popup** — anchored directly beneath the widget (left-aligned), braille-wave history graphs (U+2800+, absolute 0–100 scale), one compact mirrored block for net up/down, settings button.
- **Settings panel** — toggle modules, pick colors, refresh-rate slider, reset to defaults. Persists to `localStorage` under `hwx-monitor-config`.
- **Backend** — reuses the existing authenticated `GET /api/system/health` endpoint; no backend changes.

## Audit fixes (conventions + LOC)

- **Sync `extension.json` version** → `0.8.1` (lockstep with runtime `HermesSystemMonitor.version` + SM.02).
- **Disk IO history scale** `10485760` → `104857600` (100 MB/s = 100%) — popup graph now matches the titlebar square; was 10× too hot.
- **Extract shared `_encodeBraille()`** — removed the duplicated 12-line bitmap→braille pack loop from `wave()` and `waveNet()`.
- **Drop dead `color` param** from `wave()` + call site.
- **`closePopup()` / `closeSettings()` helpers** — collapsed 5 repeated teardown blocks; `destroy()` now detaches the outside-click listeners (was a minor leak).
- **Remove dead CSS** — `.monitor-popup-net`, `.monitor-popup-bar-wrap`, `.monitor-popup-bar`; replaced fragile fixed label/val widths with flex layout.
- **Accessibility** — added `:focus-visible` ring on the titlebar canvas for keyboard users.

## Verification

- `node --check assets/monitor.js` — OK
- `node scripts/validate-extensions.mjs` — `ok system-monitor` (18/18 entries pass)
- Playwright `specs/system-monitor.spec.ts` — **13/13 passed** (SM.01–SM.13, incl. canvas pixel scans + braille bit decode + IO format lock)

Fork PR for reference: https://github.com/ChonSong/hermes-webui-extensions/pull/11

### DIFF
--- a/extensions/system-monitor/README.md
+++ b/extensions/system-monitor/README.md
@@ -1,0 +1,90 @@
+# System Monitor
+
+Inline system resource monitor for the Hermes WebUI titlebar. Shows CPU,
+RAM, Disk, and Network usage as compact curved bars with percentage
+labels. Clicking the widget opens an expanded popup with sparkline
+history graphs. A settings panel lets you toggle modules, pick colors,
+and set the refresh rate.
+
+## What It Does
+
+- Embeds a horizontal resource monitor directly inside the `.app-titlebar`
+  strip (between the app title and the action buttons).
+- Polls `GET /api/system/health` at a configurable interval and renders
+  curved bars with a sine-wave overlay for each enabled metric.
+- Clicking the widget opens a popup showing sparkline history graphs
+  for each metric.
+- A Settings button in the popup opens a configuration panel.
+
+## Controls
+
+- **Widget (titlebar):** click to open the expanded popup.
+- **Popup:** shows sparkline history for CPU, RAM, Disk, Network.
+  - `✕` closes the popup.
+  - `Settings` opens the settings panel.
+- **Settings panel:**
+  - Toggle CPU / RAM / Disk / Network on/off.
+  - Pick a color for each module.
+  - Refresh rate slider (0.5s – 5s).
+  - Reset Defaults button.
+
+All settings persist to `localStorage` under `hwx-monitor-config`.
+
+## Trust
+
+This is a manifest-bundled client-side extension. It does not talk to
+any external origin and does not execute any sidecar or native host
+process. It only calls the already-existing authenticated
+`GET /api/system/health` endpoint that ships with Hermes WebUI.
+
+Current disclosed behavior:
+
+- reads `GET /api/system/health` (authenticated, same-origin) at the
+  configured interval.
+- writes extension-owned `localStorage` key `hwx-monitor-config`.
+- injects a `<div>` + `<canvas>` into `.app-titlebar-inner` and two
+  popup `<div>`s into `document.body`.
+- does not read any guarded WebUI globals.
+- does not need external network access.
+- does not need arbitrary filesystem access.
+- does not serve bundled assets beyond its own script/css.
+
+## Compatibility
+
+Required WebUI surface:
+
+- manifest-bundled extension assets
+- same-origin extension asset serving under `/extensions/`
+- authenticated `GET /api/system/health` endpoint
+- `.app-titlebar-inner` DOM slot for widget injection
+
+Uses only Canvas 2D and `fetch` — no third-party dependencies.
+
+## Verification
+
+```bash
+cd /path/to/hermes-webui
+HERMES_WEBUI_EXTENSION_DIR=/path/to/hermes-webui-extensions/extensions/system-monitor \
+HERMES_WEBUI_EXTENSION_MANIFEST=manifest.json \
+./start.sh
+```
+
+Manual verification:
+
+1. Titlebar shows `[CPU ██░░ 32%] [RAM ████░ 58%] ...` inline.
+2. Numbers update every refresh interval (default 2s).
+3. Click the widget → popup with sparkline graphs appears below titlebar.
+4. Click `Settings` → panel with toggles, color pickers, refresh slider.
+5. Toggle CPU off → CPU module disappears from titlebar and popup.
+6. Change refresh rate → updates happen at the new interval.
+7. Reload page → settings persist.
+8. Click outside popup → popup closes.
+9. Zero console errors on page load (`node --check assets/monitor.js`).
+
+## Known Limitations
+
+- Stats come from the backend `/api/system/health` endpoint — if that
+  endpoint is unavailable, the widget shows `—%` and a small warning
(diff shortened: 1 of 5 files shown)


nesquena-hermes:
## Review — full gate (Codex security/breakage + my own reproduction against real Core)

Thanks for this, @chonsong — a titlebar CPU/RAM monitor with curved bars and a braille-wave popup is a genuinely nice idea, and the extension.json/manifest/skin conventions are all in order. But the full gate turned up a few blockers I need addressed before it can go in. Two of them are load-bearing (a real XSS sink and modules with no data source), so this is a bounce with concrete fixes rather than a merge.

### Must-fix

**1. Stored/self-XSS: config values are interpolated into `innerHTML`.** (`assets/monitor.js` — popup rows ~205/212, settings rows ~251/256)
Color config (`cfg.cpuColor`, `cfg.netUpColor`, …) is read from `localStorage` and written straight into markup, e.g.
```js
d.innerHTML = `...<span ... style="color:${colorUp}">${txVal}</span>...`;
r.innerHTML = `<label>${l}</label><input type="color" value="${cfg[k]}">`;
```
A persisted value like `red">` becomes executable markup. `HermesSystemMonitor.setConfig()` is a second unvalidated path into the same config. Fix: validate every field on read against a strict schema (hex color regex for colors, numeric range for `refreshMs`), and build the dynamic nodes with `document.createElement` + `textContent` + `.style.color = ...` / `.value = ...` instead of HTML-string interpolation. Drop any record that fails validation.

**2. Disk IO and Network modules have no data source — they render zero/unavailable forever.** (`assets/monitor.js:92` reads `/api/system/health`)
I checked the Core endpoint you poll (`api/system_health.py` → `build_system_health_payload`, at the pinned/current head): the payload contains only `cpu`, `memory`, and `disk` aggregates. There is **no** `disk.io` / IO-rate field and **no** `net`/network object. So the "IO" and "Net ↑/↓" bars, their history, their settings toggles, and the README/metadata that advertise them can never show real data. Fix: either remove the Disk-IO and Network modules (and their settings, history, README claims, and metadata) until a supported endpoint provides those fields, or add/propose the endpoint first. Advertising a metric the app can't supply is the main thing blocking this.

**3. Permission surface understates authority.** (`extension.json:20`)
`webui_api.read: []` is declared empty, but the code performs an authenticated `GET /api/system/health`. Set `"webui_api": { "read": ["system/health"], "write": [] }`. (The safety scan passes today only because its endpoint allowlist doesn't cover this route — the disclosure is still required.)

**4. Route-relative API call can 404 on session routes.** (`assets/monitor.js:92`)
You go through Core's `api('/api/system/health')` helper, which strips the leading slash and resolves against `document.baseURI` — on a `/session/<id>` route that can become `/session/api/system/health` → 404 → the widget silently shows nothing. Use an absolute `fetch('/api/system/health', { credentials: 'same-origin' })`, check `response.ok`, and parse JSON. (This also makes any "doesn't read guarded WebUI globals" README claim true.)

**5. Poll teardown is race-prone.** (`assets/monitor.js:264/276`)
`destroy()` clears the established interval, but the initial `fetchStats().then(() => { draw(); startPoll(); })` can resolve *after* `destroy()` and resurrect polling; the `init()` retry (`setTimeout(init, 300)` when the host is absent) also retries forever, and async interval callbacks can overlap. Track a `destroyed` flag and all pending timers/requests, guard post-fetch work behind it, cancel the init retry on destroy, and prefer a completion-scheduled `setTimeout` loop over `setInterval` so polls can't overlap.

### Note on process
This is a new **visible titlebar surface** (chat-adjacent), so once the code is clean it also needs a UX pass (Fable) + populated screenshots at real desktop/mobile viewports and maintainer sign-off before merge — that's our standard flow for any visible UI, not a comment on your work. Get #1–#4 in first and I'll re-gate, then we can do the visual round.

Nice groundwork here — the CPU/RAM/Disk path is solid; it's really the two phantom modules + the XSS sink that need to land before this is mergeable.

*(Reviewed at head `1e6262b2f`. Codex security/breakage gate + my own read against real Core `320789ae`. No merge/label action taken.)*

franksong2702:
## Maintainer intent check — response requested by 2026-09-06

This head has not changed since the existing review, and #87 has now shipped the adjacent `sysinfo` host-monitoring extension. Are you planning to continue this PR?

If yes, please first state the titlebar widget value that remains distinct from `sysinfo`, then refresh onto current `main` and address the existing blockers: strict config validation/no dynamic `innerHTML`, removal of unsupported IO/network metrics unless a real data source exists, honest API disclosure, absolute same-origin fetch, and teardown-safe polling. A populated desktop/mobile UX pass remains required for the visible titlebar surface.

If there is no response or new head by 2026-09-06, we will close this as inactive queue housekeeping. That would not reject a narrower successor with a clear non-overlapping purpose.

nesquena-hermes:
## Closing as inactive queue housekeeping — thanks @ChonSong

Following the maintainer intent-check above (response requested by 2026-09-06): the head hasn't changed since the existing review, there's been no reply, and in the meantime #87 shipped the adjacent `sysinfo` host-monitoring extension. So I'm closing this for now as queue housekeeping — not a rejection of the concept.

The door stays open, and specifically:
- A **narrower titlebar-widget successor with a clear, non-overlapping purpose** distinct from `sysinfo` is welcome — the first thing to state up front would be what unique value the titlebar surface adds beyond the shipped `sysinfo` panel.
- If you take it forward, the existing blockers are the path to merge: strict config validation with no dynamic `innerHTML`, removal of the unsupported IO/network metrics unless there's a real data source, honest API disclosure, absolute same-origin fetch, teardown-safe polling, and a populated desktop/mobile UX pass for the visible titlebar surface.

Reopen or open a fresh PR anytime and I'll pick the review right back up. Thanks for the contribution.

PR #79, opened 2026-08-09
feat(typography): add role-based font controls

starship-s:
## Summary

- Add a **Typography** rail editor for choosing interface, conversation, and code fonts through Core's `--font-ui`, `--font-conversation`, and `--font-mono` tokens.
- Provide curated role-specific Google Fonts, curated presets, live previews, and selected-family CSS loading.
- Support browser-local `.woff2`, `.woff`, `.ttf`, and `.otf` imports with IndexedDB persistence, preview, rename, replacement, and deletion.
- Bound local storage to 8 fonts, 10 MiB per file, and 40 MiB total. Startup uses cursor-based bounded loading and preserves over-cap records without loading or deleting them.
- Add focused Node checks and an isolated real-browser lifecycle smoke without packaging test code in the install artifact.

## Trust and compatibility

Google-hosted selections request only the selected families and expose the browser IP to Google. The editor discloses this behavior. WebUI defaults make no Typography font request. Local font bytes stay in the browser and are never uploaded, synchronized, or sent to Google.

The extension has no backend, sidecar, native host, remote script, filesystem access, cookies, or WebUI API access. Its runtime manifest grants only an extension-owned browser-storage namespace for scalar font selections; font bytes remain in IndexedDB. Deletion fails closed when browser confirmation is unavailable.

Typography requires the merged three-font Core contract from [nesquena/hermes-webui#5918](https://github.com/nesquena/hermes-webui/pull/5918). Compatibility CI pins Core release `exp-v0.52.184` at `51e3d09ad677a6b7e4ec3db46611b383d67a2778`, the first release that includes this contract.

## Screenshots

These screenshots came from extension head `f9e0d6ec` in an isolated standalone WebUI instance at Core `51e3d09a`, with fresh temporary state and only Typography installed.

### Desktop

Typography Nous preset with Courier Prime

### Narrow/mobile

Typography narrow mobile panel

## Verification

- Extension validation and safety scanning pass for all 18 entries.
- Typography syntax, rail-entry, catalog, preset, capacity-boundary, local-font validation, selection-persistence, and Google CSS checks pass.
- All 16 compatibility unit tests pass.
- The existing maintained-pin `mobile-conversations` browser smoke passes.
- The new Typography-only real-browser smoke passes all five isolated cases:
  - default token and no-egress behavior;
  - Google stylesheet creation, update, and removal;
  - import, IndexedDB persistence, reload activation, replacement, deletion, and `FontFace` cleanup;
[...]

PR #87, opened 2026-08-19
sysinfo extension (relaunch of #67): host-control hardening + update-target parity

asorourx:
Relaunch of #67 (withdrawn) — adds the **sysinfo** extension (system Insights card: speed test + Docker stats/controls/image-updates via a bundled loopback sidecar) with the full host-control review addressed. Rebuilt on current `main` and re-verified.

## Addresses the exact-head review

**@franksong2702 — final-review blockers (all fixed)**
1. **Hermetic tests wired into CI.** `test_op_ownership.py` set `HERMES_SYSINFO_STATE_DIR` to a temp dir *before* importing `docker_stats` (its module-level epoch init used to write the real `~/.hermes/webui`); both regression files now run in `extensions.yml`.
2. **Per-container stop/restart confirm.** `mcDockerAction()` routes `stop`/`restart` through `_mcConfirmDestructive()` (fail-closed if the host has no dialog), matching the stack path. `start` stays unprompted.
3. **Unknown pull result is no longer "already latest".** `docker_update()` is tri-state (`True`/`False`/`None`): "no change" only when equality was *established*; an unknown pre-pull digest returns `changed=None` + `changed_known=false` + an honest note. Bulk result and the toast carry the tri-state through.
4. **Bounded remote manifest cache.** `_remote_cache_put()` caps entries (`MC_DOCKER_REMOTE_CACHE_MAX`, default 512): evict TTL-expired first, then oldest.
5. **Count/label/action parity + naming.** The header counts only compose-managed updatable containers and reads **"Update stacks (N)"**; the "Ungrouped" bucket gets a non-actionable badge (no stack button → no `missing_project`); standalone rows are informational only. "dependency-first" is dropped everywhere (it's a name/image heuristic, not a Compose graph).

**@nesquena-hermes — full-QA items**
- `--no-deps` / hidden-replica authorization, Compose-file reconstructibility, bulk-reservation cleanup — covered (regressions included).
- **Deadline mismatch** — both browser pollers now follow the server (no fixed ~16-min cap), so a live host mutation is never abandoned.
- **Host-scale memory** — remote cache bounded (above) + inventory collection stops at a bounded working set (`_MAX_SCAN`, default 2000) instead of materializing the full list; truncation reported.
- **localStorage** — one bounded, pruned group-state key (legacy per-project keys migrated).
- **Consent + posture** — fail-closed gate now also rejects `posture === 'local_unprotected'`, and names the password→approve remedy.
- **`extension.json`** — `/docker/updates` in both `read` and `write`; no stray `mc.docker.updates` key.
- **README** — deny-by-default default documented; heuristic ordering, not "dependency-first".
- **Kebab a11y** (the non-blocking follow-up) — ArrowUp/ArrowDown/Home/End roving focus added to the `role="menu"`.

## Tests
Sidecar regressions (`test_op_ownership.py`, `test_update_target_parity.py`) run by path in CI: op/bulk ownership, replica auth, compose reconstructibility, missing-old-digest → unknown, established-equal → "latest", bounded cache eviction, bounded inventory scan, and count/label/action parity. `node --check` on the assets.

@franksong2702 @nesquena-hermes — ready for another exact-head review; thanks for the thorough host-control pass on #67.


franksong2702 (changes requested):
Thanks for the substantial work here. I reviewed the exact head `24e5d877fc8880a3e7266be703a4e4a6e60c308d`. Two blockers need to be resolved before the maintainer review can continue.

1. **Please rebase this branch onto current `main` (`3f464a3ca2e9fdb77ebf18e107f3665e6e985a18`) and rerun the full gates.** The merge base is still `afe51f414710bef7e8bd816263a08268924ef958`; this head is 25 commits on its side and is missing seven commits now on `main` (#79, #80, #81, #82, #83, #84, and #86). The current green checks therefore do not certify the System Info extension together with today's validator, E0/B1/Configure baseline, and Typography changes. This is an integration blocker, not a request to treat the misleading 41-file two-dot comparison as this PR's diff; the actual PR diff remains the 19 files from the merge base.

2. **Roll back background-job reservations when `Thread.start()` fails, and cover both sibling paths with tests.** In `sidecar/sysinfo.py`, `_start_speedtest()` sets `_st_running = True` before `thread.start()` (lines 120–129). If thread creation raises, the request fails but `_st_running` remains true; every later manual or scheduled speed test is then rejected until the sidecar restarts. The same class exists in `sidecar/docker_stats.py` (lines 500–504): `_updates_sweeping["on"]` is set before `Thread.start()`, so a start failure permanently leaves update checks reporting `sweeping: true`. The Docker mutation paths in this PR already demonstrate the required shape: catch the start failure, release the owned reservation, and return/report a fail-closed error. Please add deterministic injected-start-failure tests for both speed-test and update-sweep paths.

Reproduction evidence for the second point (no real speed test or Docker operation was run):

```text
speed test: first_error=thread-start-injected, stuck_running=True, second_start=False
update sweep: first_error=thread-start-injected, stuck_sweeping=True, second_refresh_sweeping=True
```

The hermetic operation-ownership tests, update-target parity test, JavaScript syntax check, validators, safety scan, sidecar contract check, scaffold sync check, and sidecar-usage check passed on this exact head. I did not start the sidecar or invoke real Docker. The generic scaffold subprocess test could not be completed on this macOS/Python 3.14 host because its one-second health startup probe timed out; please rely on the post-rebase Ubuntu CI run for that gate.

nesquena-hermes:
## Re-gate at `e373204` — **SAFE TO SHIP** after two fix rounds on-branch (ready to merge pending final look)

Thanks @asorourx — this is the highest-privilege class we ship (loopback sidecar with Docker host control), so it got the full treatment: Codex security/breakage reproduce-gate + Fable UX gate, plus my own line-by-line read of the sidecar auth boundary, the docker mutation authorization, and the frontend, and the loopback-sidecar security checklist. The relaunch cleanly closes everything that bounced #67 — the token-v1 auth boundary, the docker allowlist/hidden-replica/`--no-deps` authorization, the job+poll long-op model, and all five of the final #67 exact-head blockers.

Rather than bounce for the remaining items, I took them on-branch (co-authored to you) and looped the gates to convergence.

### Round 1 — Codex `SHIP-WITH-FIXES` (6) + Fable `SHIP-WITH-UX-FIXES` (5), all applied
**Codex (reliability / disclosure / test-quality — no security defect):**
1. `sidecar/sysinfo.py` `_start_speedtest()` — roll back `_st_running`/`_st_run_thread` + set `_st_last_error` on `Thread.start()` failure (was wedging every later speed test until restart). New regression, mutation-verified.
2. `sidecar/docker_stats.py` `docker_updates()` — roll back `_updates_sweeping` on sweep-thread start failure (was leaving `sweeping:true` forever). New regression, mutation-verified.
3. `extension.json` — `/docker/updates` added to `webui_api.write` (`?refresh=1` starts work + persists).
4. `assets/sysinfo.js` `mcRunSpeedtest` — poll while the server reports `running`, bounding only *consecutive* transport failures (the fixed 90s deadline was shorter than the 120s backend timeout → false "timed out").
5. `sidecar/test_op_ownership.py` — force a fresh temp `HERMES_SYSINFO_STATE_DIR` (was `setdefault`, which gave no isolation).
6. `sidecar/test_update_target_parity.py` — real unchanged-default-state oracle (the old path-string compare passed vacuously); mutation-verified.

**Fable (UX):** `.mc-docker-startall` → `var(--success)` (was `var(--accent)`, a brand color that rendered Start/Stop both red under some skins); kebab menu item colors → `var(--success/--warning/--error)` with hex fallbacks (AA contrast on the light menu surface); reduced-motion now also zeroes transitions (the speed-test bar fills via a width transition); ≥24px tap targets on rename / group-update / kebab / update-pill; `aria-modal="true"` on the auto-schedule dialog.

### Round 2 — the re-gate caught two second-order defects my own round-1 fixes introduced, both fixed
- **Codex `SHIP-WITH-FIXES`:** my sweep-rollback cleared `sweeping:false` **without a terminal error state**, so the frontend would stamp *"all up to date · just now"* on a check that never ran. Added `_updates_sweep_error` (set on thread-start failure, unavailable inventory, and worker exceptions; cleared when a new sweep starts; returned as `sweep_error`); `mcDockerCheckUpdates` now surfaces it as a real failure. Strengthened the rollback regression to assert the surfaced error (mutation-verified).
- **Fable `SHIP-WITH-UX-FIXES`:** the reduced-motion kill matched elements only, so the `.mc-docker-row--busy::after` spinner kept rotating; extended to `*::before, *::after`. Also `.mc-docker-stopall` `var(--danger)` → `var(--error)` for theme parity (core defines `--error`, not `--danger`).

### Convergence — both gates clean on the same head `e373204`
- **Codex: SAFE TO SHIP** — *"Findings: none."* Central token gate rejects direct unauthorized mutations with 401/503; docker mutations enforce filtered-inventory membership + validated IDs/refs + workdir containment + `--no-deps` + hidden-replica checks; long ops are background job+poll; `/health` has the required CORS/no-store; no auto-consent / off-origin fetch / credential leak / dynamic-code sink / unescaped HTML; disclosure honest; scaffold byte-identical.
- **Fable: SHIP-UX** — both round-2 fixes resolved, nothing new, no crown-jewel surface touched (Insights secondary surface only).
- **CI green** on `e373204`: Validate extensions + Browser extension compatibility both pass, including *"Test sysinfo sidecar regressions (op/bulk ownership + update-target parity): success"* — the regression suite (now with the speed-test + sweep rollback tests) actually ran and passed on the runner.

**One honest gap** (unchanged from your own note, and disclosed): live Docker-daemon integration (an actual `compose pull`/recreate against a real daemon) was not exercised. The authorization *logic* around it is fully covered by the fail-closed enumeration + filtered-inventory checks + the regression suite, but the container recreate itself is untested here.

Since this is a host-control sidecar extension **and it changed in response to review this round**, I'm **not auto-merging** — @franksong2702 @rodboev, one last look when you have a moment? If you're both good, it's ready to merge.

*(Re-gated at `e373204`. Codex reproduce-gate SAFE TO SHIP + Fable SHIP-UX + my own reads + CI validators + the sidecar regression suite green. All fix commits co-authored to @asorourx. No merge/label action taken.)*

asorourx:
Hey @franksong2702 could you please give this a look with get a chance ?

franksong2702:
Thanks for the ping, @asorourx. I reviewed the current state at `e373204`. Nathan’s re-gate and the green extension checks give good confidence that the reliability fixes landed, including the `Thread.start()` rollback paths.

One integration blocker from my formal review still remains: this head is currently 8 commits behind `main@9a37dc81` (the branch has diverged). Because this extension controls a privileged Docker sidecar, please rebase `sysinfo-v2` onto current `main` and let both required checks rerun. Once that exact rebased head is green, please request review again and I’ll do the final integration-focused pass.

I’m leaving the existing `changes-requested` / `needs-rebase` disposition in place until that happens.

franksong2702 (approved):
Maintainer integration re-gate at `ff33e734`: approved. I rebased the existing 27 commits onto `main@9a37dc81`; `git range-diff` reports all 27 patches unchanged, the branch is now behind 0, the full local extension/sysinfo gates passed, and both GitHub checks passed on this exact head. The previously disclosed live Docker-daemon integration gap remains unchanged and is not introduced by the rebase.

PR #96, opened 2026-08-31
Add Custom User Avatar extension (user-turn avatars, V2 — closes #63)

### LINKED ISSUE #63: Wishlist: Custom Avatar V2 — optional user avatars + settings (from core #2586)
## Source and current state

This wishlist carries forward the remaining gap from [`nesquena/hermes-webui#2586`](https://github.com/nesquena/hermes-webui/issues/2586), which requested custom avatars for user and assistant turns.

The library now ships two assistant-side avatar entries:

- [`custom-avatar`](https://github.com/hermes-webui/hermes-webui-extensions/tree/main/extensions/custom-avatar), a single custom assistant avatar;
- [`profile-avatars`](https://github.com/hermes-webui/hermes-webui-extensions/tree/main/extensions/profile-avatars), added through #66, amended through #69, and hardened through #70, with per-profile avatar storage and rendering.

PR #70 has merged, hardening `profile-avatars` upload, storage, and renamed-root identity behavior. It does not add a user-turn avatar.

The unique unmet scope is therefore **optional user-turn decoration**. A future implementation must coexist with the two shipped entries and must not introduce a third competing assistant-avatar storage or rendering path.

Related retrofit tracker: #34.

## Why this stays an extension

Hermes WebUI gives user turns no avatar slot: right alignment identifies the speaker. Adding portraits to every turn is a personal visual preference and costs horizontal space, especially on mobile. It should remain opt-in.

## Remaining V2 scope: optional user avatar

Add an extension setting, disabled by default. When enabled, the extension should:

1. create an extension-owned avatar node for each user turn;
2. position it consistently beside the right-aligned user bubble without changing message semantics;
3. decorate current and newly rendered turns through a bounded, idempotent observer path;
4. hide or compact at narrow mobile breakpoints where an avatar would reduce readable width;
5. remove every injected node, listener, style, and observer when disabled or after reload/uninstall.

Core has no user-role avatar element, so this is an explicitly unstable DOM-mutation contract. Scope selectors to real visible user rows and fail harmlessly when transcript markup changes.

## Settings and storage

Expose extension-owned controls for:

- user avatar image / clear;
- enable user avatars;
- avatar size within a small bounded range;
- mobile behavior: hide or compact.

Requirements:

- keep the user image in extension-owned storage only;
- normalize to a small raster thumbnail before storing;
- accept PNG/JPEG/WebP; reject SVG and oversized/invalid files;
- no remote image fetching through core;
- no cross-device sync claim;
- do not write profile, personality, memory, or core settings;
- do not read, rewrite, migrate, or shadow `custom-avatar` or `profile-avatars` storage.

## UX requirements

- Avatars are decoration only; do not add bordered message cards or change content semantics.
- Preserve text width and alignment as much as possible.
- Mobile chat remains readable at 390 px and narrower.
- The no-user-avatar fallback remains pixel-identical when disabled.
- The image does not imply model/personality identity beyond the user's explicit choice.
- Settings and upload controls have keyboard labels and correct focus behavior.

## Acceptance checklist

- [ ] Existing `custom-avatar` and `profile-avatars` behavior/storage remain untouched.
- [ ] User avatars are disabled by default.
- [ ] Enabling decorates every current/new visible user turn without changing message content.
- [ ] Hidden anchors/worklog nodes are never decorated.
- [ ] Desktop and mobile layouts remain readable; narrow layout follows the configured hide/compact policy.
- [ ] Invalid, oversized, and scriptable image types are rejected.
- [ ] Images never leave the browser or enter core/profile storage.
- [ ] Repeated rerenders do not duplicate nodes or trigger observer loops.
- [ ] Disable/reload/uninstall removes injected UI and observers.
- [ ] Manifest/safety validation plus realistic mobile/desktop browser QA passes.

## Attribution

Original user+assistant avatar request: @kinower in `nesquena/hermes-webui#2586`.

Existing `custom-avatar` implementation: @nesquena-hermes. Existing `profile-avatars` implementation: @asorourx. Preserve all relevant credit in a future V2 PR.

asorourx:
## Custom User Avatar — user-turn avatars (V2), closes #63

Adds `user-avatar`, the **user-turn** complement to `custom-avatar` (assistant) and
independent of `profile-avatars`. It's the unmet V2 scope from #63: an optional,
**disabled-by-default** avatar beside each right-aligned user bubble. It does not add a
third assistant-avatar path and does not read, write, or shadow either existing
extension's storage.

### Rendering contract — no injected child nodes

Core rebuilds a user row's `innerHTML` whenever it changes and **skips** unchanged rows.
Injecting a child node into a row would defeat that skip (rebuild every render) and be
wiped on the next reconcile. So this extension sets **only element-level state core does
not reconcile**:

- a per-row `data-hwx-uav` attribute (survives `innerHTML` rebuilds and row recycling);
- `--hwx-uav-img` / `--hwx-uav-size` custom properties + a `data-hwx-uav-on` flag on `:root`.

The avatar is drawn by a `::before` **pseudo-element**, gated on `:root[data-hwx-uav-on]`.
Because the decoration is attribute/CSS-driven, the `MutationObserver` only marks newly
created user rows and a core rebuild leaves decoration intact (idempotent, no observer
loops). User rows carry `content-visibility:auto`; its paint containment clips
outside-the-box drawing, so the avatar renders **inside** the row and the bubble yields a
little width for it (the acknowledged horizontal cost).

### Storage & settings

- Scalars (`enabled`/`size`/`mobile`) via `settings_schema` + the sanctioned
  `window.hermesExt` scoped settings, with a localStorage fallback on older core
  (mobile-haptics pattern).
- The image blob lives in extension-owned `localStorage` (`hermes-ext-user-avatar`),
  downscaled to 96×96 in-browser. `permissions.storage.owned: true`.
- No network, no sidecar, no core/profile/personality/memory writes.

### Acceptance checklist (#63)

- [x] `custom-avatar` / `profile-avatars` behavior/storage untouched (separate keys; no shared reads/writes)
- [x] Disabled by default (`settings_schema` default `false`; `:root[data-hwx-uav-on]` gate)
- [x] Enabling decorates current + newly rendered user turns without changing content
- [x] Hidden anchors/worklog nodes never decorated (selector scoped to `.msg-row[data-role="user"]`)
- [x] Narrow layout follows the hide/compact policy (`@media (max-width:600px)`)
- [x] Invalid / oversized / scriptable image types rejected (validated `data:image/(png|jpeg|gif|webp)` only; SVG rejected)
- [x] Images never leave the browser or enter core/profile storage
- [x] Repeated rerenders don't duplicate nodes or loop the observer (attribute-driven, childList-only observer)
- [x] Disable/reload/uninstall removes injected UI + observers (`teardown()`; disabled = pixel-identical to stock)
- [x] Manifest/safety validation passes
- [ ] **Realistic mobile/desktop browser QA + screenshots** — still to add (I can't run a browser here; screenshots pending)

### Verification (all green)

```
node scripts/validate-extensions.mjs      # ok user-avatar
node scripts/scan-extension-safety.mjs     # safety scan passed
node scripts/test-user-avatar.mjs          # behavior tests (wired into extensions.yml)
node scripts/generate-registry.mjs --out dist/registry.json
```

`scripts/test-user-avatar.mjs` runs the real asset in a `node:vm` fake DOM and asserts:
disabled-by-default, enable/size/mobile/image application, **no child-node injection**,
idempotency, teardown, and the localStorage fallback path.

### Still needs a human

Realistic desktop + 390px screenshots and touch/scroll QA (per the repo's usual ship
gate) — happy to add screenshots once a maintainer confirms the placement/size choices,
or adjust the left-of-bubble placement if you'd prefer it beside the right edge.

Credit: original request @kinower (`nesquena/hermes-webui#2586`); wishlist + V2 scope @nesquena-hermes.


nesquena-hermes:
## Exact-head review at `bf2ac95e7` — changes needed across Configure lifecycle, storage authority, media bounds, and visual proof

The pseudo-element direction is sensible: it avoids injecting children into Core message rows, preserves the row innerHTML skip, scopes decoration to user rows, gates the feature off by default, and does not touch either assistant-avatar store. The exact generic checks are green.

The implementation still has several load-bearing defects.

### 1. Configure uses the wrong Core contract and can remain pending forever

Core invokes a registered Configure handler with one **options object**:

```js
{ opener, restoreFocus }
```

It does not pass a DOM host (`static/extension_settings.js:490-536`). The returned promise owns pending state and must settle when the extension-owned editor closes.

`user-avatar.js:350-355` instead treats the options object as a possible DOM host, falls back to `document.body`, appends an unframed panel, and waits for a MutationObserver to see `!host.contains(panel)`. `host` is the options object and has no `contains`; that callback throws on mutation, the promise never resolves, and the Core Configure button can remain disabled as “Opening…” indefinitely.

Use the shipped Theme Creator/Typography pattern:

- accept the options object intentionally;
- open an owned modal/overlay with dialog semantics;
- expose explicit X/Escape/backdrop close paths;
- trap Tab while open;
- settle the Configure promise exactly once on close/teardown;
- let Core restore opener focus once (do not compete with it);
- suppress duplicate Configure while pending.

Add a production-shaped test through `HermesExtensionSettings._invokeConfigure`, including close, failure, teardown, and second-open reuse.

### 2. Image storage bypasses the sanctioned extension namespace

`window.hermesExt.register('user-avatar')` returns both scoped `settings` and scoped `storage`, but the image reads/writes raw `localStorage['hermes-ext-user-avatar']` (`:28-34,97-110`). Current Core’s Settings → Clear extension storage clears the namespaced `ext.storage` record, not that raw key. Uninstall/clear therefore leaves the avatar behind despite `permissions.storage.owned:true`.

Use `scoped.storage.get/set/remove('image')` when available. If an old-Core fallback is retained, migrate the legacy raw key once into scoped storage and remove it after successful publication. Tests must drive the real Core clear-storage path and prove the image/decoration disappear.

### 3. Media limits are bypassable and the accepted format exceeds issue scope

The file-picker path has an 8 MiB byte cap and decodes through ``, but:

- there is no decoded dimension/pixel cap before canvas work; a small compressed image with enormous dimensions can exhaust browser memory. Reuse the existing Profile Avatars boundary: max 4096 px per side and 16 megapixels before drawing.
- public `setImage(dataUrl)` bypasses file size, decode, dimensions, and downscale entirely. `isDataImage()` checks only syntax and `setImage()` has no `MAX_BYTES` guard, so a huge raw data URL or animated GIF can be stored/applied directly.
- issue #63 explicitly accepts PNG/JPEG/WebP; this PR adds GIF. Either remove GIF or obtain an explicit issue-scope decision. Canonicalized first-frame GIF support is not implied by the tracker.
- storage/quota failure is swallowed, but Configure still reports “Saved.” Return a real success/error result and preserve the previous image on publication failure.

Prefer one private canonicalization path that always outputs bounded PNG/JPEG, and expose no raw synchronous setter that bypasses it.

### 4. Teardown is not complete

A pending `requestAnimationFrame(schedule)` is not canceled. Because the enabled setting remains true, its callback can run after teardown and re-add `data-hwx-uav` markers. `unmarkRows()` also selects only rows currently marked as `data-role="user"`; a recycled row whose role changed can retain the extension attribute.

On teardown:

- cancel and zero pending RAF;
- remove `[data-hwx-uav]` from **all** elements, independent of current role;
- close/settle any Configure UI;
- disconnect all observers/listeners/timers;
- make repeated teardown inert;
- prove no later callback re-applies state.

### 5. Current tests do not exercise these boundaries

`scripts/test-user-avatar.mjs` never invokes the Configure hook, image decode/downscale, FileReader/canvas, storage quota failure, observer callback, role recycling, or a post-teardown queued RAF. It validates only the shallow fake-DOM state setters.

Add discriminating behavior tests for every item above. Mutation-bite the Configure settlement, storage clear, media cap, and teardown guards.

### 6. Visible gate is missing

The PR explicitly has no screenshots. Before merge, provide realistic populated desktop and 390 px mobile proof for:

- disabled pixel parity;
- Small/Medium/Large;
- mobile Hide and Compact;
- long user messages, files/code blocks, hover actions, and several consecutive turns;
- default and accessibility font sizes;
- light/dark and representative skins;
- Configure focus/keyboard/error states.

Run Fable on those final screenshots. The exact placement cannot be approved from CSS alone because user rows are `align-self:flex-end` with responsive max widths and mobile `content-visibility` containment.

Threat scan is **CLEAN**, but no contributor code was bare-executed locally; hosted checks are supplemental evidence only.

**Verdict: changes requested; cross-layer and below the bounded maintainer-fix threshold.** No branch mutation or merge action from this lane.

nesquena-hermes:
## Exact-head review at `bf2ac95e7` — changes needed across Configure lifecycle, storage authority, media bounds, and visual proof

The pseudo-element direction is sensible: it avoids injecting children into Core message rows, preserves the row innerHTML skip, scopes decoration to user rows, gates the feature off by default, and does not touch either assistant-avatar store. The exact generic checks are green.

The implementation still has several load-bearing defects.

### 1. Configure uses the wrong Core contract and can remain pending forever

Core invokes a registered Configure handler with one **options object**:

```js
{ opener, restoreFocus }
```

It does not pass a DOM host (`static/extension_settings.js:490-536`). The returned promise owns pending state and must settle when the extension-owned editor closes.

`user-avatar.js:350-355` instead treats the options object as a possible DOM host, falls back to `document.body`, appends an unframed panel, and waits for a MutationObserver to see `!host.contains(panel)`. `host` is the options object and has no `contains`; that callback throws on mutation, the promise never resolves, and the Core Configure button can remain disabled as “Opening…” indefinitely.

Use the shipped Theme Creator/Typography pattern:

- accept the options object intentionally;
- open an owned modal/overlay with dialog semantics;
- expose explicit X/Escape/backdrop close paths;
- trap Tab while open;
- settle the Configure promise exactly once on close/teardown;
- let Core restore opener focus once (do not compete with it);
- suppress duplicate Configure while pending.

Add a production-shaped test through `HermesExtensionSettings._invokeConfigure`, including close, failure, teardown, and second-open reuse.

### 2. Image storage bypasses the sanctioned extension namespace

`window.hermesExt.register('user-avatar')` returns both scoped `settings` and scoped `storage`, but the image reads/writes raw `localStorage['hermes-ext-user-avatar']` (`:28-34,97-110`). Current Core’s Settings → Clear extension storage clears the namespaced `ext.storage` record, not that raw key. Uninstall/clear therefore leaves the avatar behind despite `permissions.storage.owned:true`.

Use `scoped.storage.get/set/remove('image')` when available. If an old-Core fallback is retained, migrate the legacy raw key once into scoped storage and remove it after successful publication. Tests must drive the real Core clear-storage path and prove the image/decoration disappear.

### 3. Media limits are bypassable and the accepted format exceeds issue scope

The file-picker path has an 8 MiB byte cap and decodes through ``, but:

- there is no decoded dimension/pixel cap before canvas work; a small compressed image with enormous dimensions can exhaust browser memory. Reuse the existing Profile Avatars boundary: max 4096 px per side and 16 megapixels before drawing.
- public `setImage(dataUrl)` bypasses file size, decode, dimensions, and downscale entirely. `isDataImage()` checks only syntax and `setImage()` has no `MAX_BYTES` guard, so a huge raw data URL or animated GIF can be stored/applied directly.
- issue #63 explicitly accepts PNG/JPEG/WebP; this PR adds GIF. Either remove GIF or obtain an explicit issue-scope decision. Canonicalized first-frame GIF support is not implied by the tracker.
- storage/quota failure is swallowed, but Configure still reports “Saved.” Return a real success/error result and preserve the previous image on publication failure.

Prefer one private canonicalization path that always outputs bounded PNG/JPEG, and expose no raw synchronous setter that bypasses it.

### 4. Teardown is not complete

A pending `requestAnimationFrame(schedule)` is not canceled. Because the enabled setting remains true, its callback can run after teardown and re-add `data-hwx-uav` markers. `unmarkRows()` also selects only rows currently marked as `data-role="user"`; a recycled row whose role changed can retain the extension attribute.

On teardown:

- cancel and zero pending RAF;
- remove `[data-hwx-uav]` from **all** elements, independent of current role;
- close/settle any Configure UI;
- disconnect all observers/listeners/timers;
- make repeated teardown inert;
- prove no later callback re-applies state.

### 5. Current tests do not exercise these boundaries

`scripts/test-user-avatar.mjs` never invokes the Configure hook, image decode/downscale, FileReader/canvas, storage quota failure, observer callback, role recycling, or a post-teardown queued RAF. It validates only the shallow fake-DOM state setters.

Add discriminating behavior tests for every item above. Mutation-bite the Configure settlement, storage clear, media cap, and teardown guards.

### 6. Visible gate is missing

The PR explicitly has no screenshots. Before merge, provide realistic populated desktop and 390 px mobile proof for:

- disabled pixel parity;
- Small/Medium/Large;
- mobile Hide and Compact;
- long user messages, files/code blocks, hover actions, and several consecutive turns;
- default and accessibility font sizes;
- light/dark and representative skins;
- Configure focus/keyboard/error states.

Run Fable on those final screenshots. The exact placement cannot be approved from CSS alone because user rows are `align-self:flex-end` with responsive max widths and mobile `content-visibility` containment.

Threat scan is **CLEAN**, but no contributor code was bare-executed locally; hosted checks are supplemental evidence only.

**Verdict: changes requested; cross-layer and below the bounded maintainer-fix threshold.** No branch mutation or merge action from this lane.

nesquena-hermes:
## Re-gate at `9aa1d7c748f8` — all five source-level items closed; only the visible-proof gate (screenshots + Fable) remains

Thanks @asorourx. I re-read the full extension (`user-avatar.js`, 618 lines) and the behavior suite (`test-user-avatar.mjs`, 718 lines) at the exact head, and verified each prior finding against the code — not the commit message.

### Prior findings 1–5 — closed, with non-vacuous tests

1. **Configure now uses Core's real contract and always settles.** `registerConfigure()` passes `settings.registerConfigure(() => Promise)`; the handler no longer treats the options object as a DOM host. `openPanel()`/`buildPanel()` mount an extension-owned overlay with `role="dialog"` + `aria-modal`, and `settleConfigure()` resolves exactly once. The test drives a faithful mirror of `HermesExtensionSettings._invokeConfigure` (frozen `{opener, restoreFocus}`, settle-once, Core-restores-opener-focus, duplicate-suppress-while-pending) and asserts: X/Escape/backdrop all close+settle once, a click *inside* the card does not, the focus trap wraps both edges and excludes the hidden file input, a second open after close settles again cleanly, **and a Configure that can't mount (`document.body = null`) settles immediately rather than hanging Core in "Opening…".** That last one directly bites the original never-resolves defect.

2. **Image lives in the sanctioned scoped namespace.** Reads/writes go through `scoped.storage.get/set/remove('image')`; `migrateLegacyImage()` moves the pre-0.2.0 raw key in once and deletes it only after a successful scoped write. Tests prove Core's Settings→Clear-extension-storage (`extStore.clear()`) removes both the image and the decoration, that the raw key is never written when scoped storage exists, and — importantly — that a **refused** migration keeps the legacy value rather than losing the image.

3. **Media bounds are enforced and the raw setter is gone.** `canonicalize()` checks type (PNG/JPEG/WebP; GIF deliberately dropped per issue #63 scope), `MAX_SOURCE_BYTES` (8 MB), and decoded side/pixel caps **before** any canvas work; `publishImage()` is the only writer and returns a real `{ok,error}`. The public API exposes only `setImageFile` (async, canonicalizing) and `clearImage` — no raw synchronous `setImage`. Tests assert `canvasDraws.length === 0` for oversized-bytes / over-4096px / over-16MP / zero-dimension sources (so a small file that decodes huge cannot reach the canvas), that the canvas output itself is byte-capped, and that a storage-failure reports the real error and **preserves the previous image** instead of saying "Saved."

4. **Teardown is complete and post-teardown callbacks are inert.** `teardown()` sets `installed=false`, cancels the pending RAF, disconnects the observer, clears the reconcile interval, removes the storage listener, closes+settles Configure, and `unmarkRows()` strips `[data-hwx-uav]` from **every** carrier regardless of current role. Tests queue a frame via the observer, tear down, and confirm the cancelled frame (and a force-flushed one) re-marks nothing; a recycled row that changed role is unmarked by `pruneRows()`; and a repeated teardown is inert.

5. **Tests exercise all of the above** in a real `node:vm` sandbox with a faithful fake DOM, scoped-storage record with `false`-on-fail semantics, and canvas/FileReader/Image mocks. They are discriminating, not shape-only.

CI is green (`Validate extensions` + `Browser extension compatibility`) and the PR is `MERGEABLE`.

### Finding 6 — the one remaining blocker: visible proof

`extensions/user-avatar/screenshots/` still contains only `.gitkeep`. This is a visible chat-surface change (avatar drawn by a `::before` on right-aligned user rows, responsive max-widths, mobile `content-visibility` containment), so per the project's screenshot-quality gate it can't be approved for merge from code/CSS alone. Before this is merge-ready it needs realistic **populated** desktop + 390 px mobile proof — disabled pixel parity, Small/Medium/Large, mobile Hide/Compact, long user messages with files/code blocks and hover actions across several consecutive turns, default + accessibility font sizes, light/dark + a representative skin, and the Configure focus/keyboard/error states — with a Fable pass on the finals.

**Verdict: source-clean at `9aa1d7c748f8`; merge gated only on the finding-6 visible proof.** Once realistic screenshots are attached I'll do the pixel-level inspection pass. No local contributor-code execution and no branch/merge mutation from this review lane.

nesquena-hermes:
## Re-gate at `3357879fb467` — code still clean and screenshots landed, but the screenshots reveal a real avatar-placement bug (desktop + compact)

Thanks @asorourx — the visual gate, the smoke test, and the 18 screenshots are exactly what finding 6 needed, and inspecting them was worth it: they surface a placement defect that code review alone could not.

### Code delta since my last re-gate — clean

The delta `9aa1d7c748f8..3357879fb467` is one commit and does **not touch `user-avatar.js`** — the extension logic I verified source-clean is byte-identical, so findings 1–5 still stand. The new material is `manifest.json` (+49), `extension.json` (+40/-6), README, the `tests/compatibility/user_avatar_smoke.py` real-browser gate (+830), workflow wiring, and the screenshots. `extension.json` permissions are correctly minimal and honestly declared (`webui_api.read/write` empty, `dom.owned:true`+`mutates_core_views:true`, `storage.owned:true`, no network/sidecar/native-host), and the Configure modal renders cleanly (dialog semantics, preview, upload/remove, toggle, both selects, close X — all well-aligned).

### The screenshots show the avatar is not seated next to the bubble (blocking)

I inspected the PNGs at pixel level. `mobile-hide` is correct (no avatar) and the Configure modal is clean, but the core decoration is misplaced on desktop and in compact:

- **`desktop-medium` and `desktop-large`:** the avatar floats **detached in blank space far to the left of its right-aligned bubble**, separated by the full empty gutter, and is **not top-aligned** with the bubble. On `desktop-medium` the topmost avatar is **clipped by the viewport top edge**.
- **`mobile-compact`:** the avatar **overlaps the top-left corner of the bubble** instead of resting beside it.

The PR's own demo conversation states the intent — "keep the avatar pinned to the top-left of the row, and never let the reserved gutter collapse" — and the screenshots show it isn't achieving that.

### Root cause (from `user-avatar.css`)

The `::before` is `position:absolute; left:0` **on `.msg-row`**, with space reserved by `.msg-body { margin-left: size + gap }`. But user rows are `align-self:flex-end` with a `max-width`, so the **row box is wider than the bubble** and the bubble sits at its right end. `left:0` therefore anchors the avatar to the far-left edge of the full-width row — the entire empty gutter away from the bubble — while the `.msg-body margin-left` only shifts the bubble's inner content, it doesn't pull the avatar over. `top:10px` is a fixed guess that doesn't track the actual bubble top (hence the vertical mis-alignment), and in compact the same anchoring puts the shrunk avatar over the bubble corner.

Suggested direction (your call on the exact mechanism):
- Anchor the avatar to the **bubble**, not the row — e.g. position the `::before` on `.msg-body` (which is the right-aligned element) with a negative `left: calc(-1 * (size + gap))`, keeping the `margin-left` reservation so text never sits under it; or make the row a shrink-to-fit flex end-aligned container so `left:0` actually coincides with the bubble's left edge.
- Replace the fixed `top:10px` with an anchor that tracks the bubble's first line (align to `.msg-body`'s top rather than the row's padding box).
- Re-check the compact case once anchored to the bubble; the overlap should resolve with the avatar off the bubble's left edge.
- Re-capture the desktop + compact shots after the fix so the avatar reads as a neatly seated, top-aligned gutter avatar, and confirm the topmost row isn't clipped.

### Verdict

Source-clean at `3357879fb467`, screenshot gate now satisfied for *coverage* but **not for quality**: the avatar placement is a real visible defect on desktop and compact, so this is **changes-needed** on the CSS anchoring, not merge-ready. Everything else (logic, permissions, tests, Configure UI, hide mode) is in good shape — this is the last mile. No local contributor-code execution and no branch/merge mutation from this review lane; the merge remains the maintainers' call once the placement is fixed and re-shot.

nesquena-hermes:
## Correction: my placement finding was a FALSE POSITIVE. Plus four maintainer rounds — two real fixes, and a detour I had to undo.

@asorourx — I owe you a correction and a fairly candid account of what happened.

### 1. My last review was wrong about the avatar placement

I said the avatar floated "detached in blank space far to the left of its right-aligned bubble," not top-aligned, clipped on `desktop-medium`, and overlapping the bubble corner in `mobile-compact`. **That was a false positive.** Your CSS was correct. I'm sorry — that's a frustrating thing to be told to fix when nothing is broken.

Two independent measurements, because I didn't want to correct one unverified claim with another:

**Your own committed screenshots, measured at pixel level** (isolating the avatar circle and the bubble fill, not eyeballing):

| Screenshot | Avatar x-range | Bubble starts | Gap | Vertical | Overlap |
|---|---|---|---|---|---|
| `desktop-medium` | 831 → 863 | 874 | **11px** | top-aligned | none |
| `mobile-compact` | — | — | correct | — | **none** |

**An independent probe** built from Core's real `static/style.css` and real user-row markup: gap **10px**, vertical delta **-2px**, no overlap, no clipping.

The mechanism, which is what I got wrong: user rows are `align-self:flex-end` inside a column flex container, so the row box is **shrink-to-fit around the bubble** — measured `rowWidth=228.8` against a `780` container. `left:0` therefore lands on the gutter your `margin-left` reserves, exactly as you intended. I had assumed the row was full-width. It isn't.

Your `content-visibility` rationale in the comments was also off (it only applies under `@media (pointer: coarse)`), but the *code* was right for a different reason. I've corrected the comments to describe the real contract.

### 2. Two real defects the screenshots couldn't show

Both reproduced against Core's real markup before I touched anything, both now fixed on your branch:

- **Attachments strip:** Core emits `.msg-files` *before* `.msg-body` in the user row and zeroes its left margin, so a 3-thumbnail strip extends further left than the bubble — the avatar painted **32px over the first thumbnail**. Now reserves the same gutter.
- **Edit mode:** Core makes the row full-width and sets the edit textarea's left margin to `--msg-rail` (0), so the avatar covered **32px of the textarea**. Decoration is hidden while editing and the gutter released so the editor keeps full width.

Probe: 2 collisions → 0, with plain-row placement byte-identical before and after (avatar left/top and the 10px gap unchanged at every viewport).

- **Configure ignored write failures.** The toggle and both selects didn't check `writeScalar()`'s result, so a refused write left the control showing a value that was never stored or applied. They now revert to the authoritative value and report it. This one was pre-existing, and worth having.

### 3. The part where I made it worse, then undid it

In my first round I "fixed" `writeScalar()` to stop writing the raw localStorage key when the scoped write succeeded — I read your dual write as a bug. **It wasn't, and I should have asked before changing it.**

That change orphaned the raw key, which forced a `migrateLegacyScalars()` in round 2 to repair the damage, which then needed conflict resolution, retry state and multi-tab race handling — none of which your original design needed. Round 3's gate found three further ordering breaks in that machinery. At that point the signal was unmistakable: the premise was wrong, not the implementation. I reverted the whole detour.

Round 4 then caught that a *second* change of mine (`return ok || rawOk`) defeated the very Configure guard I'd added in the same commit. Also reverted, back to your original `return ok || !settingsSupported`, which is the correct rule.

Net effect of four rounds: **your `writeScalar` is exactly as you wrote it.** Everything I "fixed" about it, I put back.

One thing the gate did surface that is genuinely worth knowing: the raw mirror is **one-way and unversioned** — Core's native Save/Reset writes the scoped store directly without passing through `writeScalar()`, so the mirror can go stale, and Reset settings doesn't clear it. That only bites on a downgrade to a Core without scoped settings. I documented the limitation in the JS comment and README rather than building a versioned migration protocol inside this PR. If we want true downgrade fidelity it deserves its own change.

### 4. What's on your branch now

Four commits, `Co-authored-by: asorourx` on each — authorship stays yours:

- `721bb20` — CSS collision fixes + disclosure corrections
- `e206833` — *(the migration detour — reverted below)*
- `f49c6cc` — reverts the detour; keeps the Configure failure guard
- `cc5c7fd` — restores the scoped-authoritative write result; corrects the mirror disclosure

Every behavioral fix is mutation-verified: the suite goes red for `return ok || rawOk`, and for the toggle, size select and narrow-screen select each ignoring a failure. The harness now models Core's real schema-default behavior instead of returning `undefined`, and the failure test uses the realistic scoped-refused / raw-succeeds case — requiring *both* stores to fail is what let my round-3 bug through in the first place.

Gates at `cc5c7fd`: validate-extensions 20/20, scan-extension-safety passed, behavior tests 12/12, collision probe 0. CI green (Validate extensions + Browser extension compatibility). `MERGEABLE`.

### Where this stands

I'm **not** merging it. I've now had four rounds on this branch and each one found a defect in my own previous round, which is exactly the point at which I should stop making changes rather than start a fifth. Every remaining item is closed as far as I can verify, but the last word on your own extension should be yours.

Two things I'd like from you:

1. **Sanity-check my edits**, particularly that the revert genuinely restored your intent in `writeScalar` and that the collision fixes read correctly to you.
2. **Re-shoot `desktop-medium` and `mobile-compact` with attachments and edit mode in frame**, if you can — those two states are the ones the original 18 screenshots didn't cover, and they're where the real bugs were hiding. The existing shots are fine for everything else.

Then I'm happy to merge, or you can push over anything of mine you disagree with. The extension itself is good work — a clean pseudo-element approach that survives Core's innerHTML reconcile, honest permissions, and a real-browser visual gate is more rigor than most extension PRs arrive with.

_Gate: Codex (gpt-5.6-sol) across four rounds plus Fable UX; my own real-fixture collision probe, pixel measurement of the committed screenshots, and a flex-layout mechanism probe. Final head `cc5c7fdc29e235e72ee76c89880909f1a6f767c2`._

nesquena-hermes:
## Correction to my own note: `settings.overrides` **is** available, so the migration limitation is solvable

@asorourx — following up on my previous comment with a correction, because I got a technical detail wrong in it and it changes what's actually possible here.

I wrote that resolving the one-way mirror properly "needs a versioned migration protocol," and in an earlier round I'd concluded that a conflict-aware migration wasn't implementable because there was no way to tell "the user explicitly chose this" from "Core handed back a schema default." That was wrong. I re-read `static/extension_settings.js` at the pinned Core and:

```js
// settingsAccessor() — the same accessor an extension receives from
// window.hermesExt.register(), static/extension_settings.js:288
get overrides(){ return currentOverrides(); },
```

`overrides` returns **only the keys explicitly set**, not the schema-default-filled `values` map. That is exactly the missing signal. `settings.get(key)` answering a default for an unset key is what defeated my round-2 migration; `settings.overrides` distinguishes the two cleanly, and it's a public getter on the accessor extensions already hold — not an internal.

So the honest position is: the stale-mirror limitation I documented in the code comment and README **is** solvable, with roughly this shape:

- on install, for each scalar, migrate the raw key only when `!(key in settings.overrides)` — an existing user choice always wins;
- delete the raw key only after the scoped write is confirmed;
- keep raw writes for the `!settingsSupported` case only, so a modern Core stops writing a shadow that can go stale.

I am deliberately **not** implementing that in this PR. I've already made four rounds of changes to your branch, three of which I had to revert, and the right move is to stop rather than attempt a fifth redesign of your persistence layer on my own judgment. It's also genuinely your call: the current dual-write behaves correctly on every supported path, and the limitation only shows up on a downgrade to a Core without scoped settings, which may not be worth the complexity.

What I'd suggest: leave this PR as it stands and treat the migration as a separate follow-up, either from you or from me with your sign-off. If you'd rather fold it in now, the sketch above is what I'd write, and I'm happy to review it.

What changes in this PR as a result: nothing in the code. But the comment in `user-avatar.js` and the README both currently say the limitation needs "a versioned migration protocol," which overstates it — `settings.overrides` makes it a modest, bounded change. I'll correct that wording if you want it precise now, or fold it into whichever change actually implements the migration.

Everything else from my previous comment stands: gates green at `cc5c7fd`, CI green, `MERGEABLE`, and I'm still not merging — waiting on your sanity-check of my revert and, if you can manage it, re-shot `desktop-medium` / `mobile-compact` with attachments and edit mode in frame.

_Correction based on re-reading `static/extension_settings.js:264-300` at the CI-pinned Core, after a gate round flagged the same API and I initially dismissed it._

nesquena-hermes:
## Maintainer finish: merged with three fix commits on top of `cc5c7fd`

@asorourx, thank you for pushing this through several rounds. The extension is solid. The final Codex and Fable rounds found only narrow issues, so rather than bounce it again I finished them on the branch. Your commits are preserved as-is (merge commit, no squash).

### What changed on top of your head

**`e899980`: persistence has a single authoritative store**
- Before, the mirror model wrote scoped settings and the raw `hermes-ext-user-avatar-*` localStorage keys side by side, and a stale raw key could mask the scoped value. Now there is one store per Core generation: scoped settings when available, raw keys only on older Cores.
- `migrateLegacyScalars()` folds raw values in once, one way. A non-default scoped value wins, junk values are ignored, and a refused adoption leaves the raw key untouched.
- The older-Core fallback returns the real write result, so a refused write reverts the control instead of reporting success.
- CSS: avatar `top` 10px → 12px, level with Core's real user-row padding. The close button is 44×44 under `(pointer: coarse)`.
- README: rewrote the persistence disclosure to match.

**`8f9ceb4`: attachment and edit-mode coverage, plus a precise disclosure**
- Your smoke now has two layout cases (desktop Hide, mobile Compact). Each renders a user turn with three real PNG attachments through Core's own `/api/file/raw` URL shape, then enters edit mode through Core's `editMessage()`. It asserts there is no overlap between the avatar and the strip, thumbnails, bubble, edit textarea or edit bar. The avatar hides only on the row being edited and returns on Cancel. Four new screenshots.

**`3ed5c2b`: teardown, inline-form sync, stacking, exact stub**
- `teardown()` now calls the disposer Core's `registerConfigure()` returns, and the handler is inert once uninstalled (it settles immediately, opens nothing).
- **Inline-form sync.** Core also renders these settings as the inline "Browser-local extension settings" form, and its Save writes every field. Without the sync, a size picked in Configure could be silently reverted by a later inline Save. After each Configure write the extension now refreshes its own inputs in those forms. The smoke drives this end to end.
- Overlay z-index 70 → 1000, matching profile-avatars. The smoke asserts the modal is topmost at the backdrop and card centre.
- The attachment stub matches only the exact origin, route, seeded session id and filename, and requires each file exactly once.
- `extension.json` lists the new screenshots, and `docs/compatibility-smoke.md` documents the new checks.

### Evidence
Against current Core `master` (`30b407439`):
- Codex convergence gate on `3ed5c2b`: **SAFE TO SHIP**, no findings.
- Fable UX gate on `3ed5c2b`: **SHIP-UX**.
- `test-user-avatar` green, with all 10 targeted mutations killed.
- Real-Core smoke: 18 cases passed, and every new smoke check was mutation-verified. CI's "Browser extension compatibility" job also passed on `3ed5c2b`.
- `validate-extensions`, 12/12 behavior suites, `scan-extension-safety`, sidecar checks, `validate-desktop-companion` and `generate-registry` are all green.

One deliberate non-change: Fable suggested Small as the default size. I kept your Medium default. It's a product preference, and Configure makes Small one click away.

Optional follow-up from Fable, not gating: `flex-wrap: wrap` on the Configure preview and setting rows as insurance below the tested 390px floor (320px phones).

cc @franksong2702 @rodboev for a post-merge look, since maintainer commits landed on top of the contributor head.

Closes #63.

franksong2702:
Post-merge recheck of the three maintainer follow-ups, on merged extension tree `9e08e3bd5e13ca3ff5479d5bf6263f0e6b116be1` against an independent clean Core checkout at `65b6e4d390ec93036cab05b2f9e3cd6a817c376b`.

I traced the scalar migration against Core's actual settings accessor: default-equal values are indeed removed from overrides, and the documented one-time legacy/default ambiguity matches that contract. Non-default scoped values win, refused adoption retains the raw key, and Configure writes synchronize this extension's inline fields. The Configure disposer and inert post-teardown callback also match the current host contract.

Fresh verification:
- user-avatar behavior suite passed; four targeted temporary-source mutations were all detected (overwrite scoped choice, discard refused migration, omit Configure disposer, claim failed legacy write succeeded);
- all 15 registered behavior suites passed; extension validation, safety scan and both sidecar checks passed;
- real-Core Chromium visual/Configure track: 18/18 cases passed, including inline Save, modal stacking, attachments/edit mode and disabled PNG parity at desktop and 390px;
- inspected the resulting geometry/parity/network evidence: no unexpected egress; tracked source unchanged.

No new blocker found in the maintainer delta. I have not changed the default size, CI pin, or the optional sub-390px layout behavior. This used disposable state and synthetic transcripts, not a live Agent/provider or physical-device acceptance. AI-assisted review with GPT-6 Astra Pro and WebCodex; no independent human sign-off is implied. No new code, merge, release or deployment was performed.

PR #98, opened 2026-09-15
Add opencode-go-usage extension: OpenCode Go plan-usage chip + loopback sidecar

Kioraga:
## OpenCode Go Usage

Adds the **opencode-go-usage** extension: a passive **OpenCode Go** status chip in
the composer footer (right after `.composer-divider`) that opens a panel with the
OpenCode Go plan usage — the plan's own live windows (rolling 5 h / weekly /
monthly) as percentages with their reset times — straight from OpenCode's
documented `GET /zen/go/v1/usage` endpoint. Once data arrives the chip shows the
three percentages as `Go: x·y·z%`.

### How it works
- A `manifest-bundle` frontend (`assets/opencode-go-usage.js` + `.css`) that adds a
  chip to the composer footer and a right-aligned panel with a callout tail,
  mirroring the core chip and context-window tooltip styles. No core views are
  modified.
- A `loopback-sidecar` on `127.0.0.1:17799` that holds the Go API key, calls
  OpenCode's usage endpoint, and returns the plan windows through the WebUI's
  consented `token-v1` sidecar proxy. The browser never sees the key.
- Hover opens the panel (like the context-window indicator); it stays pinned
  until `Escape` / an outside click / clicking the chip again. Auto-refresh and
  the interval are configurable in the extension settings.

### Trust and disclosure
- One external egress only: the sidecar's single outbound `GET
  opencode.ai/zen/go/v1/usage` with the user's `OPENCODE_GO_API_KEY`. The browser
  assets contact no external origin.
- The sidecar reads two local files, both read-only: the key from `~/.hermes/.env`
  (when not already in its environment) and `~/.hermes/state.db` (`mode=ro`,
  `session_model_usage`). It writes nothing anywhere.
- The `token-v1` proxy is consent-gated and fail-closed; WebUI authentication is
  required. The sidecar binds to loopback and its only tokenless route is `/health`.
- No secrets, tokens, credentials, or machine-specific state are committed.

### Verification
- `node scripts/validate-extensions.mjs`, `scripts/scan-extension-safety.mjs`,
  `scripts/sync-sidecar-base.mjs --check`, `scripts/check-sidecar-usage.mjs` all
  pass locally.
- Sidecar gate: tokenless probe → `503`, bad token → `401`, valid token → `200`
  with real plan windows; `/health` is the only unauthenticated route.
- A DOM smoke test exercises the composer insertion, hover open/pin, refresh
  (the panel keeps its size, the refresh glyph spins), and per-window rendering.

Tested against WebUI with `token-v1` sidecar auth and `extension-settings`
(`HermesExtensionSettings.settingsForExtension`).

### Notes
- The extension id/folder is `opencode-go-usage` (matches the display name).
- A follow-up can register the DOM smoke test in `scripts/behavior-tests.json`;
  it currently lives out of tree.


nesquena-hermes:
## Review at `248e19e1b` — strong sidecar engineering; four blockers + a UX placement call

Thanks @Kioraga — this is one of the better-built sidecar extensions we've had. Before the findings, credit where it's due, because these are exactly the things that usually go wrong and didn't:

- **The token-v1 boundary is correct.** Every non-`/health` route is authenticated at the scaffold's dispatch chokepoint. No "only the proxy can reach me" hand-wave.
- **No auto-consent.** The frontend never POSTs `approved:true`; it detects the un-approved state and tells the operator where to approve it. `diagnose()` mapping 403/`local_unprotected` → "WebUI authentication is off" is genuinely good operator UX.
- **Both protected scaffold files are byte-identical** to `examples/sidecar-scaffold/` (I diffed them), the systemd `ExecStart` is the pinned `/usr/bin/python3 -S -u sidecar.py` form, `WorkingDirectory` resolves to the declared runtime path, and the runtime tree has no symlinks or import-shadowing artifacts.
- **Zero XSS surface.** Every sidecar-controlled string goes through `textContent`/`el()`. No `innerHTML` anywhere.
- **API paths are absolute** (`/api/extensions/...`), so no route-relative 404 on `/session/<id>`.
- `validate-extensions`, `scan-extension-safety`, and `check-sidecar-usage` all pass (20 entries).

Gates: **Codex → SHIP ONLY WITH FIXES**, **Fable (UX) → UX-REWORK**. I reproduced each finding below myself rather than forwarding the verdicts.

---

### 1. 🔴 MUST-FIX — the OpenCode API key leaks across a redirect

`sidecar/opencode_usage.py:198-208` builds an `Authorization: Bearer <key>` header and calls `urllib.request.urlopen()`, which **follows redirects while preserving that header**. Any redirect — a compromised/changed endpoint, a captive portal, DNS interference — receives your OpenCode key.

I reproduced this against a local two-server harness (origin 302s to a second port):

```
Redirect target received Authorization: 'Bearer SUPER_SECRET_KEY_12345'
LEAKED
```

**Fix** — refuse redirects entirely for the credentialed request:

```python
class _NoRedirect(urllib.request.HTTPRedirectHandler):
    def redirect_request(self, req, fp, code, msg, headers, newurl):
        raise urllib.error.HTTPError(req.full_url, code, f"refusing redirect to {newurl}", headers, fp)

_OPENER = urllib.request.build_opener(_NoRedirect)
# ...
with _OPENER.open(req, timeout=_HTTP_TIMEOUT_SECONDS) as resp:
```

Please add a regression test asserting the redirect target is never contacted and never sees the key.

### 2. 🔴 MUST-FIX — `/api/usage` can exceed core's 512 KiB proxy cap → 502

`build_payload()` (`opencode_usage.py:434`) always computes `local_usage("go")` and returns it as `go.local`. Two collections in that payload are **uncapped**: `models[].tasks[]` and `unpriced_models[]`. (`models[]` itself is correctly capped at `_MAX_MODEL_ROWS = 40` — the cap just doesn't cover these two.)

I measured it by feeding a synthetic `state.db` through your own `local_usage()`:

| shape | total `go.local` bytes | dominant field |
|---|---|---|
| 20k rows, 40 models × 20k tasks | **646,818** (over cap → 502) | `tasks[]` = 94% |
| 20k rows, 20k models × 40 tasks | **644,376** (over cap → 502) | `unpriced_models[]` = 94% |

Note I could *not* reproduce a 6,000-row failure — at 6k rows I measured ~41–64 KB, comfortably inside the cap. It takes tens of thousands of rows or very high model/task cardinality. So this is a **long-lived-instance** failure, not a day-one one, which is worth knowing when you decide how to fix it.

The simplest fix is also the best one here: **the frontend never reads `go.local`.** I grepped `opencode-go-usage.js` for `go.local`/`.local` — zero references. Dropping it from the route removes the size risk, the SQLite read, and the 5s busy-timeout from the request path entirely. If you want the local-accounting view later, it wants to be its own bounded/paginated endpoint.

### 3. 🟠 MUST-FIX — `timeout=6` is not a 6-second deadline

`timeout=` on `urlopen` is a **per-socket-operation** timeout, not wall-clock, and `resp.read()` (`opencode_usage.py:209`) is unbounded. A slow-drip response can hold the request open past core's hard **10s** proxy timeout (→ 502) and read unbounded bytes into memory. Cap the body before parsing (e.g. `resp.read(_MAX_UPSTREAM_BYTES + 1)` and reject if larger), and consider a background single-flight refresh that returns cached/loading state immediately.

### 4. 🟠 MUST-FIX — "Nothing is written" is inaccurate

`extension.json:53` and `README.md:114` both claim nothing is written. Running the committed service command creates `sidecar/__pycache__/*.pyc` — I verified `-S` does **not** suppress bytecode caching:

```
$ /usr/bin/python3 -S -u main.py
__pycache__/helper.cpython-312.pyc
```

(`.gitignore` correctly excludes them, so this is a disclosure fix, not a code one.) Please reword to "no application or state data is written; Python may create bounded bytecode caches inside the extension's own runtime directory."

---

### 5. 🟠 UX — the composer-footer placement is the biggest open question

This is the one I'd like your view on before you spend effort, because it's a design call rather than a bug.

The chip mounts into `.composer-footer` right after `.composer-divider` — which puts a third-party plan readout **ahead of the YOLO pill, profile, workspace, and model chips** on our densest and most-protected surface. Concretely:

- **It shifts core's collapse threshold.** Core's `_fitComposerFooter()` measures `.composer-left` scrollWidth vs clientWidth and toggles `cf-icons` → `cf-burger`, which hides the **model and workspace text labels**. The chip is `flex-shrink:0`/`nowrap` and adds roughly 115–130px, so those labels collapse earlier on every desktop layout — and the chip has **no rule for either stage**, so it keeps full text width while core's own chips shrink to 44px icons. Core's own `#providerQuotaChip` is the precedent to follow here.
- **The async label swap re-triggers the fit.** Going from "OpenCode Go" to "Go: 11%·4%·23%" after the fetch makes labels disappear a beat after load.
- **Every other extension that adds chrome uses the titlebar** (session-export-pdf, rss-feeds, custom-branding). Your own commit `730522b` moved this *off* the titlebar — and `manifest.json` still says "Titlebar panel", so the gallery card describes a surface it no longer uses.

**Recommendation: move it back to the titlebar right cluster**, opening the same popover on click. That sidesteps the collapse interaction completely. If you feel strongly about the composer, it needs to sit after the model chip beside `#providerQuotaChip`, adopt that chip's styling, and add `cf-icons`/`cf-burger` rules.

### 6. 🟠 Smaller UX/a11y items (verified against core)

- **Undefined color tokens.** `opencode-go-usage.css:117,118,141,142` use `var(--warn, …)` and `var(--danger, …)`. Core defines **neither** — I counted 0 definitions of each, vs 23 for `--warning` and 23 for `--error`. So the bars render the hardcoded amber/red fallback in *every* theme. Switch to `--warning`/`--error`.
- **Contrast.** The live label uses `--accent` as text. In default light that measures **2.81:1** against `--surface` (AA needs 4.5). `--accent-text` measures **4.58:1** and passes. Use `--accent-text` with an `--accent` fallback.
- **The chip is a `<span>`** with no `tabindex`, role, or accessible name, and opens on hover. Keyboard and touch users can't reach it, and screen readers get nothing. Make it a `<button>` with `aria-haspopup="dialog"` + `aria-expanded`, opening on click.
- **Escape is captured and `stopPropagation`'d** on `document`, which swallows core's own Escape handling while the panel is open.
- **`fixedPanelHeight` is module-level and never reset** (`closePanel` doesn't clear it), so one tall error state makes every later open that tall until reload.
- **Docs vs code:** the README's "moving away closes it" claim doesn't match the code (there's no `mouseleave` handler), and the "Titlebar panel" description is stale.

---

### Where this leaves us

Items 1–4 are mechanical and I'd happily take them myself, but **item 5 is a placement decision that's yours to make** (and it changes the CSS you'd write for item 6), so I'd rather not rewrite your surface out from under you. If you'd prefer, say the word and I'll push 1–4 plus the token/contrast fixes to your branch and leave the placement to you.

The security boundary here is genuinely well-built — the redirect leak is a subtle `urllib` default that catches almost everyone, not a design mistake. Happy to re-gate as soon as you've pushed.

Kioraga:
Thanks for the thorough review — all blockers addressed, plus the placement call. Pushed three commits (refuse redirects/cap/drop-local; redirect regression test; composer placement + UX fixes).

**1. Redirect leak (blocker)** — a `_NoRedirect` handler now guards the credentialed request: any 3xx raises, so the redirect target is **never contacted** and the key cannot follow a redirect. Surfaced as a distinct `redirected` plan error.

**2. Payload over the 512 KiB proxy cap (blocker)** — dropped the local `state.db` accounting (`go.local`) from `/api/usage` entirely. The frontend never read it and its uncapped collections were the size risk; the payload is now plan-only, constant-size, and no longer touches `state.db`.

**3. timeout=6 / unbounded read (must-fix)** — the upstream body is capped at 64 KiB before parsing.

**4. "Nothing is written" (must-fix)** — README + `extension.json` reworded: no application/state data is written; Python may create bounded bytecode caches in the runtime dir.

**5. Placement (your call)** — kept the composer placement per the requester's decision, but moved the chip to sit right after `.composer-model-wrap`, beside `#providerQuotaChip`, adopting that chip's styling. It now hides in the `cf-icons` / `cf-burger` collapse stages (core hides the model/quota chips there too), so it never shifts the footer fit. Manifest/README no longer say "Titlebar panel".

**6. UX/a11y minor**
- `--warn/--danger` → `--warning/--error` (core defines only the latter).
- Live label → `--accent-text` (AA) instead of `--accent`.
- The chip stays a passive `<span>` for now (requester's decision); the `<button aria-haspopup/aria-expanded>` conversion is a marked follow-up.
- Escape no longer `stopPropagation`'d on `document`.
- `fixedPanelHeight` reset on close, so one tall error state doesn't condition every later open.

**Verification** — `validate-extensions`, `scan-extension-safety`, `sync-sidecar-base --check`, `check-sidecar-usage` pass; the new `scripts/test-opencode-redirect.py` (two-server redirect harness) is registered in `behavior-tests.json` and green under the shared runner; DOM smoke test green. Ready for re-gate.

Kioraga:
Small follow-up on the chip styling (commit `40dd934`): per the requester I reverted the chip to the **composer-left pill** — same padding/`--muted` text as its neighbours, no border, no green tint — and it no longer hides in the `cf-icons` / `cf-burger` collapse stages (it still hides on phones, `max-width: 640px`). Position stays right after `.composer-model-wrap`. The color-token fixes (`--accent-text`, `--warning/--error`) are kept.

nesquena-hermes:
## Re-gate at `4f5d557` — all four blockers closed; I pushed the last two fixes myself

Thanks @Kioraga — the security work here landed properly. Verified at this head:

**1. Redirect credential leak — CLOSED.** `_NoRedirect` (`opencode_usage.py:55-69`) is installed via `build_opener`, and I confirmed the opener carries no default redirect handler. I re-ran my two-server harness against your actual `_OPENER`:

```
outcome: HTTPError code=302 reason=refusing redirect to http://127.0.0.1:.../stolen
target server hits: 0
hits carrying the key: 0
RESULT: CLEAN — redirect target never contacted, key never left
```

urllib calls `redirect_request` before `parent.open`, so the target is never dialled — not "contacted then failed". 300/301/303/307/308 all map to `redirected`. Your `scripts/test-opencode-redirect.py` passes standalone and under the shared runner.

**2. `go.local` over the proxy cap — CLOSED.** `build_payload()` (`opencode_usage.py:226-241`) is plan-only. No sqlite import, no `state.db` reference anywhere in the extension, no `.local` reference in the frontend. Dropping it rather than capping it was the right call — it removed the SQLite read and the busy-timeout from the request path too.

**3. Unbounded body read — CLOSED.** `read(_MAX_UPSTREAM_BYTES + 1)` then reject (`:158-160`) is the correct shape (rejects rather than silently truncating into a JSON parse).

**4. "Nothing is written" — CLOSED.** `extension.json:78` and `README.md:111-112` now describe the bytecode caches honestly.

Security posture re-verified end to end: no `innerHTML` anywhere, every sidecar string goes through `textContent`, one credentialed egress with redirects refused and body capped, the key never appears in a payload/log/error, auth enforced at the dispatch chokepoint (`sidecar_base.py:216`) with `/health` the only tokenless path, and both scaffold files byte-identical to `examples/sidecar-scaffold`.

---

### What I pushed to your branch

The gate flagged two HIGH-VALUE items. Both were mechanical, so I took them rather than sending you another round — commits `ba50b85` and `4f5d557`, with you as co-author.

**a. Restored the `cf-icons` / `cf-burger` collapse rules** (the ones you had in `bafdb39` and dropped in `40dd934`).

I measured this against core's **real** `.composer-footer` markup extracted from `static/index.html`, core's real `style.css`, and core's real `_fitComposerFooter()`:

| | cf-icons onset | cf-burger onset |
|---|---|---|
| baseline, no chip | 376px | **266px** |
| with the collapse rules | 524px | **266px** |
| rules removed | 524px | **420px** |

Worth being precise about what this does and doesn't fix, because my first measurement was wrong and the re-gate caught it: the rules do **not** change the cf-icons onset. That first-stage shift is inherent to adding any non-shrinking element to `.composer-left`, since the fit routine measures with both stage classes stripped — core's own chips do the same. What the rules fix is the **escalation into cf-burger**, the stage where core hides the model wrap entirely. Without them the chip keeps its full pill width after core has already shrunk its own chips to 44px icons, dragging cf-burger from 266px out to 420px. With them, cf-burger returns to baseline exactly. Removing the rules reproduces the regression, so the measurement isn't vacuous.

**b. `dom.mutates_core_views: true`** — the script inserts a child into core's `.composer-left`, which is the same insertion for which session-export-pdf, rss-feeds, custom-branding and sysinfo all declare `true`. I also fixed the README Trust section, which still said the assets "never mutate core views" and so contradicted the flag (caught on re-gate of my own commit — same disclosure-honesty class as item 4).

---

### Left for you — none blocking, your call

- **`timeout=6` is per-recv, not wall-clock** (`opencode_usage.py:157`). A slow-drip upstream held `go_plan()` for ~19.5s against the nominal 6s in a harness; core's proxy 502s at 10s and the sidecar thread stays busy until the drip ends. Either loop `read1()` against a `time.monotonic()` deadline, or correct the 6s claim at `README.md:130` and `routes_impl.py:10`. Malicious-upstream only.
- **The hover-only `<span>` chip** (`opencode-go-usage.js:509-522`). I'm respecting your requester's decision, but it's unfocusable with no accessible name, so keyboard users can't open the panel — and since the click handler toggles, a user who hovers *then* clicks (the natural gesture next to click-to-open neighbours) sees the panel flash shut. `lastFocus` (`:37`, `:477`, `:413`) is dead code and `button.focus()` (`:412`) is a no-op on a span. It's about five lines to make it a `<button type="button" aria-haspopup="dialog" aria-expanded>`.
- **Orphaned panel edge** (`:426-433`, `:459-473`). If the composer collapses during the 160ms hover delay the chip's rect is all zeros, so the panel opens off-screen with its refresh timer running. Pre-existing — it's reachable through the 640px media query on `40dd934` too — but the collapse rules add a more common desktop trigger. Bail in `openPanel()` when `button.getClientRects().length === 0`, and a `ResizeObserver` on the chip to close-or-replace would close it properly.
- **Empty-state gap** (`:273-295`): `plan.available === true` with empty `windows` renders a blank body under an "Updated hh:mm" header. And `redirected` currently falls through to the generic "The quota lookup failed (redirected)" — worth a specific message saying the endpoint tried to redirect and the key was withheld, since that's a security-relevant state you went to real trouble to create.
- **Stale strings**: the service file still advertises "locally measured OpenCode Zen usage" (`opencode-go-usage-sidecar.service:2,8`) after local accounting was removed; `routes_impl.py:8` says "Both routes" for one route; `README.md:14` says "no click" while `:517` handles click. The PR description also still mentions reading `state.db` — worth updating before merge.
- Optional UX thought, purely your call: the first-stage shift could be cut from ~148px to ~26px by letting the chip shrink (`flex: 0 1 auto; min-width: 0; overflow: hidden` on the chip, ellipsis on the label) instead of `flex-shrink: 0`. Cost is the label ellipsizing across roughly a 120px band before it hides.

---

This is **ready to merge from my side** — all gates pass (`validate-extensions`, `scan-extension-safety`, `check-sidecar-usage`, `sync-sidecar-base --check`), CI is green, and the security boundary is genuinely well-built. Since the PR changed in response to my review, I'm not self-merging: @Kioraga please confirm you're happy with my two commits, and @franksong2702 @rodboev one last look when you have a moment.

The redirect leak was a subtle `urllib` default that catches almost everyone, and dropping `go.local` entirely rather than capping it was the better instinct — nice work.

_Gate: Fable (claude-fable-5-1) — round 1 at `40dd934`: SHIP WITH FIXES (1, 2); re-gate of my own patch at `ba50b85`: SHIP WITH FIXES (1), now applied in `4f5d557`. Plus my own redirect-leak harness and real-markup collapse measurements. Codex was unavailable this round (expired auth token)._

nesquena-hermes:
## Re-gate at `2d085de` — Codex leg finally ran, found a real interaction race; I fixed it and its sibling

Worth flagging up front: my previous comment said "ready to merge from my side," but its gate footer noted **Codex was unavailable that round (expired auth token)** — so that verdict rested on Fable plus my own harness. The Codex route is healthy again, so I ran the missing adversarial leg on this PR rather than shipping on a half gate. It found something real.

### Codex (gpt-5.6-sol) — security boundary independently confirmed

Everything the security work claimed held up under a second, independent adversarial pass at `4f5d557`:

- No XSS sinks — every upstream-controlled value reaches `textContent` or a bounded numeric style value.
- The API key goes only to the fixed HTTPS OpenCode endpoint. The opener carries no default redirect handler; direct tests confirm **301/302/303/307/308 all refused with zero redirect-target hits**.
- Upstream body read once at 64 KiB + 1, oversized responses rejected.
- Listener binds `127.0.0.1`; `GET/HEAD /health` is the only tokenless route; every other route passes constant-time token comparison.
- **No** alternate fetch path, **no** configurable upstream URL (so no SSRF surface), no traceback in responses, no request logging, no key-bearing payload.
- Permissions honestly disclose WebUI reads, storage, external networking, loopback use and core DOM mutation.
- Verified against core `adbb3258`: composer selectors, quota-chip adjacency, settings API and the `cf-icons`/`cf-burger` contracts all exist.

**Codex verdict: SAFE TO SHIP**, with one non-security finding below.

### The finding — and the sibling the UX re-gate then caught

**1. Hover-then-click flashed the panel shut** (`opencode-go-usage.js`). Entering the chip arms the 160 ms open timer; clicking before it fires opens the panel, then the still-pending timer re-enters `openPanel()`, takes the toggle branch, and closes what you just opened. Hover-then-click is the natural gesture beside core's click-to-open chips, so this was easy to hit.

**2. A fly-over pinned a panel nobody asked for.** Re-gating *my own* fix turned up the sibling case: `mouseenter` was bound but nothing cancelled on `mouseleave`, so moving across the footer and away still opened a pinned panel 160 ms later — needing Escape or an outside click to dismiss. That defeats the "short delay filters intent" behaviour your README describes. I reproduced it before fixing it.

Both are the same invariant: **no pending hover timer may survive a transition.**

### What I pushed (commits `5bee115`, `2d085de`, you as co-author)

- `cancelHoverTimers()` as the first statement of `openPanel()`, before the toggle branch.
- `node.addEventListener('mouseleave', cancelHoverTimers)` beside the existing `mouseenter` binding.
- `scripts/test-opencode-hover-click.mjs`, registered in `scripts/behavior-tests.json` so it runs in CI. It drives the **real** extension source in a `vm` sandbox against core's actual `.composer-footer > .composer-left > .composer-model-wrap` subtree and replays the exact gestures.
- README fix: the feature list still said the chip had "no click" while the click handler toggles the panel.

**Mutation-verified, both fixes independently** — a green suite proves nothing until it's shown to fail without the fix:

| Mutation | Result |
|---|---|
| stub out `cancelHoverTimers()` in `openPanel()` | ❌ RED — "the pending hover timer closed the panel the click just opened" |
| stub out the `mouseleave` binding | ❌ RED — "a fly-over pinned the panel" |
| both fixes present | ✅ GREEN |

CI on this head: `Validate extensions` ✅ and `Browser extension compatibility` ✅, 13/13 behavior suites (including the new one), safety scan, sidecar contract/usage, desktop-companion and registry generation all pass.

### Still yours — none blocking

Unchanged from my last comment, and I'm respecting your call on the interaction model:

- **`timeout=6` is per-recv, not wall-clock** (`opencode_usage.py:157`) — a slow-drip upstream held `go_plan()` ~19.5 s against the nominal 6 s. Either loop `read1()` against a `time.monotonic()` deadline, or correct the claim at `README.md:130` / `routes_impl.py:10`. Malicious-upstream only.
- **The chip is a `<span>`** (`:509-522`) — unfocusable, no accessible name, so keyboard users can't open the panel. `lastFocus` is dead code and `button.focus()` is a no-op on a span. ~5 lines to make it a `<button type="button" aria-haspopup="dialog" aria-expanded>`. Your requester's decision, so I've left it.
- **Orphaned panel edge** (`:426-433`) — if the composer collapses during the hover delay the chip's rect is all zeros and the panel opens off-screen with its refresh timer running. Bail in `openPanel()` when `button.getClientRects().length === 0`.
- **Empty-state gap** (`:273-295`) — `plan.available === true` with empty `windows` renders a blank body. And `redirected` falls through to the generic failure text; that state is security-relevant enough to deserve its own message saying the endpoint tried to redirect and the key was withheld.
- **Stale strings** — the service file still advertises "locally measured OpenCode Zen usage" after local accounting was removed; `routes_impl.py:8` says "Both routes" for one route. The PR description also still mentions reading `state.db`.

---

**Ready to merge from my side**, now on a complete gate rather than a partial one. Since the PR changed again in response to my review, I'm not self-merging: @Kioraga please confirm you're happy with `5bee115` and `2d085de`, and @franksong2702 @rodboev a final look when you have a moment.

The redirect-leak defence is genuinely well-built — refusing in `redirect_request` so the target is never dialled, rather than contacting-then-failing, is the correct shape and catches almost everyone out.

_Gate: Codex (gpt-5.6-sol) at `4f5d557` — SAFE TO SHIP with finding 1. Fable (claude-fable-5-1) re-gate of my own patch at `5bee115` — SHIP WITH THE HIGH-VALUE FIXES, which found finding 2, applied in `2d085de`. Plus my own mutation-verification of both fixes and the full local CI suite. Reviewed at exact head `2d085def514cdd585ef47db98627924a5ba24621`._

nesquena-hermes:
## Merged — thanks @Kioraga 🎉

Landed on `main` as `d4f39794`, with the full commit history preserved (merge commit, not squash) so your authorship stays intact.

### What cleared it this round

My previous "ready to merge" rested on a **half gate** — that round's footer noted Codex was unavailable (expired auth token), so the verdict was Fable plus my own harness. On top of that, the final two commits on this branch were *mine* (`5bee115`, `2d085de`), and no advisor had ever seen them. A maintainer patch that bypassed contributor review deserves a harsher read, not a softer one, so I re-ran the adversarial leg on the exact head before merging rather than shipping on the earlier verdict.

**Codex (gpt-5.6-sol) at `2d085def` — SAFE TO SHIP, no findings.** It specifically confirmed:

- The sole hover-timer arm (`opencode-go-usage.js:426`) is cleared and nulled by `cancelHoverTimers()`, and every transition that should cancel one does: open/toggle, `mouseleave`, Escape, outside-click, and normal teardown. No leak path left open.
- Legitimate click-to-close still works — cancelling before the `if (panel)` toggle branch doesn't break the intentional close.
- The security boundary is unchanged by my commits: no XSS sink, no dynamic execution, no off-origin exfiltration, no SSRF input, redirect refusal / 64 KiB cap / loopback bind / token enforcement all intact, permissions honestly disclosed.

I also chased the one trap most likely to bite fix 2 — whether moving the cursor from the chip **into its child `<span class="hwx-ocu-btn-label">`** fires `mouseleave` on the parent and silently kills hover-open. Rather than assert it from memory I drove real CDP pointer input across the boundary:

```
enter chip padding  → enter=1 leave=0
move INTO child     → enter=1 leave=0   ← no spurious leave, hover-open survives
move fully outside  → enter=1 leave=1   ← cancels exactly when it should
```

That's standard `mouseenter`/`mouseleave` boundary behaviour (unlike `mouseover`/`mouseout`), confirmed empirically rather than assumed.

**Both fixes mutation-verified** — a green suite proves nothing until it's shown to fail without the fix:

| Mutation | Result |
|---|---|
| neuter `cancelHoverTimers()` in `openPanel()` | ❌ RED |
| neuter the `mouseleave` binding | ❌ RED |
| both present | ✅ GREEN |

13/13 behavior suites, `Validate extensions` ✅, `Browser extension compatibility` ✅.

### Still open — yours whenever you like, none blocking

Carried over unchanged; the redirect-refusal design is good enough that none of these held up the merge:

- **`timeout=6` is per-recv, not wall-clock** (`opencode_usage.py:157`) — a slow-drip upstream held `go_plan()` ~19.5 s against a nominal 6 s. Loop `read1()` against a `time.monotonic()` deadline, or correct the claim at `README.md:130` / `routes_impl.py:10`. Malicious-upstream only.
- **The chip is a `<span>`** (`:509-522`) — unfocusable, no accessible name, so keyboard users can't open the panel; `lastFocus` is dead code and `button.focus()` is a no-op on a span. ~5 lines to make it a `<button type="button" aria-haspopup="dialog" aria-expanded>`. Left alone since the chip-not-button look was your requester's call, but it's the one I'd most like to see.
- **Orphaned-panel edge** (`:426-433`) — if the composer collapses during the hover delay the chip's rect is all zeros and the panel opens off-screen with its refresh timer running. Bail in `openPanel()` when `button.getClientRects().length === 0`.
- **Empty-state gap** (`:273-295`) — `plan.available === true` with empty `windows` renders a blank body, and `redirected` falls through to the generic failure text; that state is security-relevant enough to deserve its own message saying the endpoint tried to redirect and the key was withheld.
- **Stale strings** — the service file still advertises "locally measured OpenCode Zen usage" after local accounting was removed; `routes_impl.py:8` says "Both routes" for one route.

Happy to take any of these as a follow-up if you'd rather not.

The redirect-leak defence deserves the last word: refusing inside `redirect_request` so the target is never dialled — rather than contacting-then-failing — is the correct shape, and it catches almost everyone out. Nice work.

_Gate: Codex (gpt-5.6-sol) SAFE TO SHIP (no findings) at exact head `2d085def514cdd585ef47db98627924a5ba24621`, plus my own CDP pointer-boundary reproduction, mutation-verification of both fixes, and the full local suite. Merged at that exact head after a pre-merge head re-check._

